PR #3, opened 2016-07-25
Feature/yaml configurations

michalwronowski:
Switching service.xml and repository.xml configurations to YAML


src/main/java/com/cognifide/knotx/repository/RepositoryConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public List<RepositoryMetadata> getRepositories() {
+        return repositories;
+    }
+
+    public void setRepositories(List<RepositoryMetadata> repositories) {
+        this.repositories = repositories;
+    }
+
+    public static class RepositoryMetadata {
```
tomaszmichalak: Please check if we could have different models for different repositories. For example we add Redis repository with authentication so we need extended model with two additional fields: user and password. We do not want to extend simple remote repository so dedicated configuration and metadata is required.

src/main/java/com/cognifide/knotx/repository/RepositoryConfiguration.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return "RepositoryMetadata{" +
+                    "path='" + path + '\'' +
+                    ", serviceUrl='" + serviceUrl + '\'' +
+                    ", catalogue='" + catalogue + '\'' +
+                    ", type=" + type +
+                    '}';
```
toniedzwiedz: This custom format may be a bit difficult to maintain across classes in a consistent way in the long run. I think there's already a dependency on Apache Commons, in which case we could use `ToStringBuilder`. Just a suggestion.

src/main/java/com/cognifide/knotx/repository/RepositoryType.java
```diff
@@ -17,20 +17,44 @@
...
-    public RepositoryBuilder getRepositoryBuilder() {
-        return repositoryBuilder;
+        @Override
+        public boolean validate(RepositoryConfiguration.RepositoryMetadata metadata) {
+            return isNotEmpty(metadata.getPath(), metadata.getServiceUrl());
+        }
+    };
+
+    private static boolean isNotEmpty(String... values) {
+        return Stream.of(values).filter(StringUtils::isBlank).count() == 0;
```
toniedzwiedz: Consider using [`noneMatch`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#noneMatch-java.util.function.Predicate-) instead of `count` as the reduction.

tomaszmichalak:
Changes accepted, ready to merge.

PR #9, opened 2016-07-29
Feature/performance

pawelkoper:
jmx test that makes requests to /content/local/simple.html


Skejven:
This branch should not be merged to master.

PR #26, opened 2016-08-10
maven license plugins refactor

wiiitek:
I have removed plugin that inserts licenses as one can copy a license from existing file.
Also we have the apache-rat-plugin that checks if license headers are present in all but excluded files.

The apache-rat-plugin was removed from child modules as it's execution is inherited from parent pom.


Skejven:
Approved.

PR #27, opened 2016-08-10
service example with remote data

wiiitek:
[http://localhost:8092/content/jsonplaceholder/remote.html](http://localhost:8092/content/jsonplaceholder/remote.html)
shows sample data from mock service and real remote service on same page.

mock data folders were renamed once again.


Skejven:
Approved.

Issue #12, opened 2016-08-02
Redundant <div> added while filling template with data


michalwronowski:
Given following snippet:
```html
<script data-api-type="templating" data-call-uri="/path/to/service.json" type="text/x-handlebars-template">
    <h2>{{header}}</h2>
    <div>{{body.content}}</div>
</script>
```
the output would be:
```html
<div>
    <h2>Header value from service</h2>
    <div>Some body content from service</div>
</div>
```
Surrounding `<div>` should not be [added](https://github.com/Cognifide/knotx/blob/master/src/main/java/com/cognifide/knotx/handler/RestServiceResponseHandler.java#L91), because in edge cases page can be rendered/styled incorrectly.

Expected output:
```html
<h2>Header value from service</h2>
<div>Some body content from service</div>
```

Skejven:
Fixed in https://github.com/Cognifide/knotx/pull/35

PR #31, opened 2016-08-16
Redundant <div> added while filling template with data (#12)

dprzybyl:
---


Skejven:
Approved

Issue #33, opened 2016-08-17
Encoding problem when retrieving compressed template


wsatyla:
When (template) response is compressed (gzip). HttpClient is unable to uncompress response body.

wiiitek:
It could be nice to have some integration tests for issues ik this.
What do you think guys?

PR #34, opened 2016-08-17
try use compression added to http client options

wsatyla:
This is fix for https://github.com/Cognifide/knotx/issues/33


knotx-core/src/main/java/com/cognifide/knotx/repository/RepositoryType.java
```diff
@@ -44,7 +46,7 @@ public boolean validate(RepositoryConfiguration.RepositoryMetadata metadata) {
         public Repository<String, URI> create(RepositoryConfiguration.RepositoryMetadata metadata,
                                               KnotxVerticle verticle) {
             return RemoteRepository.of(metadata.getPath(), metadata.getDomain(), metadata.getPort(), verticle
-                    .getVertx().createHttpClient());
+                    .getVertx().createHttpClient(new HttpClientOptions().setTryUseCompression(true)));
```
Skejven: This option should be configurable.

Skejven:
Please add proper pull request message according to [Contributing Rules](https://github.com/Cognifide/knotx/blob/master/CONTRIBUTING.md#pull-requests).

knotx-core/src/main/java/com/cognifide/knotx/repository/RepositoryType.java
marcinczeczko: I dont't think we need to have it configurable. As the method says it tries to use compression. if it's not supported it still work as expected.

wsatyla:

knotx-core/knotx-repository/src/main/java/com/cognifide/knotx/repository/RepositoryConfiguration.java
```diff
@@ -149,6 +154,14 @@ public void setType(RepositoryType type) {
             this.type = type;
         }
 
+        public JsonObject getClientOptions() {
+            return clientOptions;
+        }
+
+        public void setClientOptions(JsonObject clientOptions) {
```
marcinczeczko: I don't see clientOptions are being set anywere except unit test. Looks like you missed passing it from verticle.

wsatyla: Client options are being set at 44 line of RepositoryConfiguration:
repositoryMetadata.clientOptions = item.getJsonObject("client.options", new JsonObject());
All the setters of RepositoryConfiguration.RepositoryMetadata are used in UnitTests only

pawelkoper:
you forgot to add 'client.options': to 'knotx-example-monolith.json'

Skejven:
Approved.

PR #40, opened 2016-08-26
Feature/22 provide mechanism distinguishing from multiple forms

pawelkoper:
---


knotx-core/knotx-api/src/main/java/com/cognifide/knotx/api/JsonObjectRequest.java
```diff
@@ -43,4 +43,10 @@ protected MultiMap fromJsonArray(final JsonArray array) {
 
         return map;
     }
+
+    protected String multiMapToString(MultiMap map) {
+        StringBuilder result = new StringBuilder();
+        map.names().forEach(header -> result.append(header).append("=").append(map.get(header)).append("\n"));
```
marcinczeczko: I think you can do it a little bita shorter :)
```java
return map.names().stream().reduce (new StringBuilder(), (x, y) -> x.append ("=").append(map.getHeader(y)).toString()
```

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
```diff
@@ -44,14 +47,20 @@ public ServiceEngine(Vertx vertx, TemplateEngineConfiguration serviceConfigurati
         this.configuration = serviceConfiguration;
     }
 
-    public Observable<Map<String, Object>> doServiceCall(ServiceEntry serviceEntry, MultiMap headers, HttpMethod httpMethod) {
+    public Observable<Map<String, Object>> doServiceCall(ServiceEntry serviceEntry, MultiMap headers, HttpMethod httpMethod, MultiMap formAttributes) {
```
marcinczeczko: Well... I'd refactor this method a little bit, and use the approach from vert.x itself. THe observable created here suppose to move to kind of helper class, let's call it KnotxRxHelper in which we can but a metod for doing request that returns observable, in the form as below (it's just a draft - please double check what I have written )
```
    private Observable<HttpClientResponse> request(HttpClient client, HttpMethod method, int port, String domain, String uri, Action1<HttpClientRequest> requestBuilder) {
        return Observable.create(subscriber -> {
            HttpClientRequest req = client.request(method, port, domain, uri);
            requestBuilder.call(req);
            Observable<HttpClientResponse> resp = req.toObservable();
            resp.subscribe(subscriber);
            req.end();
        });
    }
```
Last parameter is the function call so we can use it to actually compose the request appropriately. E.g. we can invoke our helper:
```
        Observable<HttpClientResponse> serviceResponse = KnotxRxHelper.request(vertx.createHttpClient(), httpMethod, serviceEntry.getPort(), serviceEntry.getDomain(), serviceEntry.getServiceUri(),
                req -> buildRequestBody(req, headers, formAttributes));
```
While buildRequestBody() method is:
```
    private void buildRequestBody(HttpClientRequest request, MultiMap headers, MultiMap formAttributes) {
        request.headers().addAll(headers);
        if (!formAttributes.isEmpty()) {
            Buffer buffer = createFormPostBody(formAttributes);
            request.headers().set("content-length", String.valueOf(buffer.length()));
            request.headers().set("content-type", "application/x-www-form-urlencoded");
            request.write(buffer);
        }
```

knotx-core/knotx-api/src/main/java/com/cognifide/knotx/api/JsonObjectRequest.java
pawelkoper: I've decided to use Guava's joiner instead

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
pawelkoper: Thanks. I've refactor this code based on your comment

knotx-core/knotx-api/src/main/java/com/cognifide/knotx/api/JsonObjectRequest.java
```diff
@@ -43,4 +47,8 @@ protected MultiMap fromJsonArray(final JsonArray array) {
 
         return map;
     }
+
+    protected String multiMapToString(MultiMap map) {
+        return Joiner.on("\n").withKeyValueSeparator("=").join((Iterable<Map.Entry<String, String>>) map.getDelegate());
```
marcinczeczko: looks like there is missing depednency to guava now as travis is failing on this while compiling

pawelkoper: There is  guava dependency in knotx-core/knotx-api/pom.xml. I will investigate why travis build is falling

pawelkoper: it works now

marcinczeczko:
PR approved. However is there any documentation update needs to be done for the changes introduced here ?

marcinczeczko:
Approved. Please merge

PR #41, opened 2016-08-29
Feature/introduce rxjava

tomaszmichalak:
---


README.md
```diff
@@ -53,9 +53,9 @@ We care a lot about speed and that is why we built it on [Vert.x](http://vertx.i
 In order to separate static content and dynamic data we introduced a Templating Engine, which merges a template obtained from the content repository and dynamic data provided by microservices using [Handlebars.js](http://handlebarsjs.com/). Here is what a template looks like:
 
 ```html
-<script data-api-type="templating" data-call-uri="/path/to/service.json" type="text/x-handlebars-template">
-    <h2>{{header}}</h2>
-    <div>{{body.content}}</div>
+<script data-api-type="templating" data-uri-dataservice="/path/to/service.json" type="text/x-handlebars-template">
```
tomaszmichalak: In description we have data-uri-post-dataservice. Please validate what is correct value? What is default http method?

tomaszmichalak: Should we specify http method here? Please add *-get-*.

README.md
```diff
@@ -64,9 +64,9 @@ The following table describes all elements and attributes used in the template.
 | Element                             | Description                                                              |
 | ----------------------------------- | ------------------------------------------------------------------------ |
 | `data-api-type="templating"`        | required for **Knot.x** to recognize the script as a template to process |
-| `data-call-uri`                     | path to a microsevice that provides the data - it will be handled by a service, as described in the [Configuration](#configuration) section |
+| `data-uri-post-dataservice`    | path to a microsevice that provides the data - it will be handled by a service, as described in the [Configuration](#configuration) section. Name of the attribute consist of following parts:<ul><li>'data-uri' : required. </li><li>post : optional. call to  microsevicervice will be made only if request method is of that type. Can be: post, get, all. </li><li>dataservice : optional. Defines the namespace name. Only placeholders with matching namespace will be filled by data coming from that service</li></ul>|
```
tomaszmichalak: Please add description for situation with two GET services: one with namespace, second without. In handlebars can I access data from second service?

README.md
Skejven: It is possible to omit method. By default _-all-_ is applied. If doc does not say about this feature, it should be updated.

README.md
```diff
@@ -133,156 +189,219 @@ http://localhost:8092/content/remote/simple.html
...
 ```
+This section configures the Knot.x Template Engine Verticle listening for requests on Vert.x event bus. The config node consists of:
+- **service.name** - name/address of the event bus to which the Repository Verticle subscribes,
+- **template.debug** - boolean flag to enable/disable rendering HTML comment entities around dynamic snippets,
+- **services** - an array of definitions of all service endpoints used by dynamic snippets.
 
 There are two groups of services defined. Each one will be handled by a different server, i.e. all service requests which match the regular expression:
-
 - `/service/mock/.*` will by handled by `localhost:3000`
 - `/service/.*` will be handled by `localhost:8080`
```
tomaszmichalak: Do we want to provide configuration for external service in example configuration? In current situation it is not required.

README.md
```diff
@@ -133,156 +189,219 @@ http://localhost:8092/content/remote/simple.html
...
+      "service.name": "template-engine",
+      "template.debug": true,
+      "services": [
+        {
+          "path": "/service/mock/.*",
+          "domain": "localhost",
+          "port": 3000
+        },
+        {
+          "path": "/service/.*",
```
tomaszmichalak: Can be removed. Please read next comment.

README.md
```diff
@@ -133,156 +189,219 @@ http://localhost:8092/content/remote/simple.html
...
+**1.3. knotxServer** section
+```json
+{
+  ...
+  "knotxServer": {
+    "config": {
+      "http.port": 8092,
+      "preserved.headers": [
+        "User-Agent",
+        "X-Solr-Core-Key",
```
tomaszmichalak: X-Solr-Core-Key, X-Language-Code can be removed from example.

knotx-core/knotx-api/src/main/java/com/cognifide/knotx/api/RepositoryResponse.java
```diff
@@ -0,0 +1,97 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.cognifide.knotx.api;
+
+import io.vertx.core.json.JsonObject;
+import rx.Observable;
+
+public class RepositoryResponse extends JsonObjectRequest {
```
tomaszmichalak: Response extends from request? :) JsonObjectRequest -> JsonObjectModel ?

knotx-core/knotx-api/src/main/java/com/cognifide/knotx/api/RepositoryResponse.java
```diff
@@ -0,0 +1,97 @@
...
+        if (success) {
+            object.put("data", data);
+        } else {
+            object.put("reason", reason);
+        }
+        return object;
+    }
+
+    @Override
+    public String toString() {
```
tomaszmichalak: We need to verify toString result format (for example RemoteResponse[aaaa=aaa]).

Issue #23, opened 2016-08-10
Provide access to response metadata in snippet scope.


Skejven:
Provide access to response metadata in snippet scope. Snippet author should be able to see `statusCode` of the response.
Put all metadata under `_response` namespace.

### snippet:
```html
<script data-api-type="templating" 
    type="text/x-handlebars-template" 
    data-uri-post-formResponse="/service/formSubmit">
  <h1>Welcome!</h1>
    
  {{#if formResponse}}
    {{#if formResponse._response.statusCode == 200}}
    <p>Form was submitted successfully.</p>
    {{else}}
    <p>Failed to save your request. Try again later.</p>
    {{/if}}
  {{else}}
    <p>Please subscribe to the newsletter:</p>
    <form method="post">
      <input type="email" name="email" />
      <input type="submit" value="Submit" />
    </form>  
  {{/if}}
</script>
```

### `/service/formSubmit` response:
```json
{
  "status": "success"
}
```

When form was submitted and `/service/formSubmit` responses with status Code `200` (OK) following markup will be rendered:
```html
  <h1>Welcome!</h1>
  <p>Form was submitted successfully.</p>
```

### `/service/formSubmit` responses with `Status Code:404 Not Found`

When form was submitted and `/service/formSubmit` responses with status Code `404` following markup will be rendered:
```html
  <h1>Welcome!</h1>
  <p>Failed to save your request. Try again later.</p>
```

PR #46, opened 2016-08-29
#23 Provide access to response metadata in snippet scope

mariuszkubis:
---


README.md
```diff
@@ -421,6 +422,25 @@ Notice: The following `data-uri` attributes
 ```
 would trigger two calls for data because of the difference in query strings, even though the path to service is the same in both.
 
+## Service response status code
+Snippet author is able to see service response status code.
```
marcinczeczko: Should be rather something like: "Service response status code can be used in snippets e.g. as condition to display messages if response is not success".

See below other comment related to the above.

README.md
```diff
@@ -421,6 +422,25 @@ Notice: The following `data-uri` attributes
...
+```html
+<script data-api-type="templating" 
+    type="text/x-handlebars-template" 
+    data-uri-post-formResponse="/service/formSubmit">
+  <h1>Welcome!</h1>
+
+  {{#if formResponse._response.statusCode == 200}}
+    <p>Form was submitted successfully.</p>
+  {{else}}
+    <p>Failed to save your request. Try again later.</p>
```
marcinczeczko: Maybe create more possible scenario of using this:
```
if statusCode == 503
  <p> Failed ... (becuase service was unavailable)
else 
 <p> success
```

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
```diff
@@ -53,12 +56,18 @@ public ServiceEngine(Vertx vertx, TemplateEngineConfiguration serviceConfigurati
                     resp.subscribe(subscriber);
                     req.end();
                 });
-        return serviceResponse.flatMap(response ->
-                Observable.just(Buffer.buffer())
-                        .mergeWith(response.toObservable())
-                        .reduce(Buffer::appendBuffer))
-                .doOnNext(this::traceServiceCall)
-                .flatMap(buffer -> Observable.just(buffer.toJsonObject().getMap()));
+        return serviceResponse.flatMap(response -> Observable.just(Buffer.buffer())
```
marcinczeczko: Please wrap
```
flatMap(response -> Observable.just(Buffer.buffer()).mergeWith(response.toObservable()).reduce(Buffer::appendBuffer))
```
into separate method, like collectBuffers or reduceBuffers. See HttpRepository#get and HttpRespository#reduceBuffers - they doing the same. It will give you more readable code here.

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
```diff
@@ -53,12 +56,18 @@ public ServiceEngine(Vertx vertx, TemplateEngineConfiguration serviceConfigurati
...
+                .mergeWith(response.toObservable())
+                .reduce(Buffer::appendBuffer)
+                .flatMap(buffer -> {
+                    JsonObject jsonObject = buffer.toJsonObject();
+                    Map<String, String> responseMap = Maps.newHashMap();
+                    responseMap.put("statusCode", String.valueOf(response.statusCode()));
+                    jsonObject.getMap().put("_response", responseMap);
```
marcinczeczko: Don't understand why Map is doing anything here while just getting key=value pair.
- Whole invocation of this flatMap() should be:
```
.flatMap(buffer -> 
      buffer.toJsonObject().put("_response", 
                                               new JsonObject().put("statusCode", response.statusCode()) //you can extract to method called e.g. processHeaders(Buffer) if you want.
).doOnNext(this::traceServiceCall);
```

Skejven:
Please add proper pull request message according to [Contributing Rules](https://github.com/Cognifide/knotx/blob/master/CONTRIBUTING.md#pull-requests).

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
mariuszkubis: We need put data directly to map `(buffer.toJsonObject().getMap())`
This `JsonObject` is provided by `vertex.io` and it keeps value type. 
In your example we will receive` key=(JsonObject)value` par and it causes a problem in handlebars parser.

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
mariuszkubis: done

marcinczeczko: I see you put it to separate method but don't see usage of this method here

knotx-core/knotx-template-engine/src/main/java/com/cognifide/knotx/engine/service/ServiceEngine.java
```diff
@@ -53,12 +54,17 @@ public ServiceEngine(Vertx vertx, TemplateEngineConfiguration serviceConfigurati
...
-                .doOnNext(this::traceServiceCall)
-                .flatMap(buffer -> Observable.just(buffer.toJsonObject().getMap()));
+        return serviceResponse.flatMap(response -> Observable.just(Buffer.buffer())
+                .mergeWith(response.toObservable())
+                .reduce(Buffer::appendBuffer)
+                .flatMap(buffer -> Observable.just(buffer.toJsonObject().getMap()))
+                .map(results -> collectBuffers(results, response)));
+    }
+
+    private Map<String, Object> collectBuffers(Map<String, Object> results, HttpClientResponse response) {
```
marcinczeczko: you missunderstood my intention. Take a look my previous comment once again. I was saying to extract this part:
```
Observable.just(Buffer.buffer()).mergeWith(response.toObservable()).reduce(Buffer::appendBuffer)
```
. 

Current this method is doing something else but not buffer reduction..

README.md
```diff
@@ -421,6 +422,25 @@ Notice: The following `data-uri` attributes
 ```
 would trigger two calls for data because of the difference in query strings, even though the path to service is the same in both.
 
+## Service response status code
```
marcinczeczko: No more full documentation in Readme.md - please move this to documentation/wiki..

marcinczeczko:
Approved

PR #66, opened 2016-09-02
Feature/50 documentation update documentation regarding p rs

mariuszkubis:
---


documentation/src/main/wiki/SnippetProcessing.md
```diff
@@ -0,0 +1,194 @@
...
+  <form method="post">
+    <input type="email" name="email" />
+    <input type="submit" value="Submit" />
+  </form>
+  <p>{{message}}</p>
+</script>
+```
+
+###Any method services calls
+When service path is marked as data-uri-**all**, GET call will be execute regardless of incoming request method type.
```
Skejven: Maybe we should mention, that calling `data-uri` without any postfix is alias for `data-uri-all`?

Skejven: Added.

Skejven:
Approved

PR #71, opened 2016-09-12
Sortpom plugin

wiiitek:
Do we want to use sortpom plugin?

Pro:
- checks for whtespaces and indendation, sorts poms

Cons:
- first line of XML is long,
- lines would be longer than 100 chars
- the plugin is strict and sometimes we know better how to organize our poms...

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -18,12 +18,14 @@
     limitations under the License.
 
 -->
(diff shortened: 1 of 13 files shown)


documentation/pom.xml
```diff
@@ -1,103 +1,103 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
wiiitek: There were windows line endings in this pom and also whitespaces inconsistencies...

wiiitek:
This pull request was updated: sortpom plugin is removed from root pom.
We can still use it occasionally even if we don't have it in pom.

Skejven:
Approved

PR #74, opened 2016-09-14
Fixed formatting - changed tab size and ident to 2 spaces

Skejven:
Fixed tab size and ident -> set to 2 spaces as originally in google style guide for Java
---


Skejven:
Self approve! :)

PR #89, opened 2016-09-30
Knot.x verticles launcher

mariuszkubis:


documentation/src/main/wiki/GettingStarted.md
```diff
@@ -129,12 +129,14 @@ This is the main configuration supplying config entries for each verticle starte
       ]
     }
   },
+  "number.of.instances": 2
```
marcinczeczko: Verticles lanucher should support all possible parameters available for DeploymentOptions object.
It means:

- [ ] config section should look as follows
```json
{
  "some-verticle" : {
     "config" : {...},
     "some-param" : "aa",
     "other-param" : "bb"
 }
```
Simple on the same level as config - those params need to be the one that DeploymentOptions object support. It's covered in documentation.
- [ ] in the starter verticle, you can simply supply our verticle json into DeploymentOptions object.
Use (DeploymentOptions#fromJson() method) that will do all the job [See here how it works](https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/DeploymentOptions.java#L103).
We must not be redunant to what Vert.x giving us out of the box.

documentation/src/main/wiki/ProductionUsage.md
```diff
@@ -24,13 +24,21 @@ The *core* module contains 3 Knot.x verticle without any sample data. Here's how
...
       ],
-      "dependencies" : {
-        "repository.address" : "template-repository",
+      "dependencies": {
+        "repository.address": "template-repository",
         "engine.address": "template-engine"
       }
-    }
+    },
+    "worker" : false,
```
marcinczeczko: Remove those deployment properties from production usage as it might suggest that's the setting should be. Instead refer to the documentation where you explain possibilities. And here simply mention that each verticle can be configured/scaled using those properties

documentation/src/main/wiki/GettingStarted.md
```diff
@@ -260,3 +264,32 @@ This section configures the Services mock used by the example application. It co
...
+    "isolationGroup": "null",
+    "ha": false,
+    "extraClasspath": [],
+    "instances": 2,
+    "isolatedClasses": []
+  }
+    ...
+}
+```
+- **worker** - deploy verticle as a worker, default false.
```
marcinczeczko: Explain here, where those properties comes from (Vert.x [documentation](http://vertx.io/docs/apidocs/io/vertx/core/DeploymentOptions.html)). Additionally, mention here that properties having default values can be omitted - (E.G.  "extraClasspath": [] can be missleading here")

knotx-example/knotx-example-monolith/src/main/java/com/cognifide/knotx/monolith/SampleApplication.java
```diff
@@ -33,11 +27,11 @@
   public void start() {
     Vertx vertx = Vertx.vertx();
 
-    vertx.deployVerticle(new MockRemoteRepositoryVerticle(), new DeploymentOptions().setConfig(config().getJsonObject("mockRepo")));
-    vertx.deployVerticle(new MockServiceVerticle(), new DeploymentOptions().setConfig(config().getJsonObject("mockService")));
-    vertx.deployVerticle(new RepositoryVerticle(), new DeploymentOptions().setConfig(config().getJsonObject("repository")));
-    vertx.deployVerticle(new TemplateEngineVerticle(), new DeploymentOptions().setConfig(config().getJsonObject("engine")));
-    vertx.deployVerticle(new KnotxServerVerticle(), new DeploymentOptions().setConfig(config().getJsonObject("server")));
+    vertx.deployVerticle("com.cognifide.knotx.mocks.MockRemoteRepositoryVerticle", new DeploymentOptions(config().getJsonObject("mockRepo")));
```
marcinczeczko: Please do helper method for deployment, e.g.
```java
private void deployVerticle(Vertx vertx, String verticleName, String configName) {
   vertx.deployVerticle(verticleName, new DeploymentOptions(config().getJsonObject(configName)));
}
```
And here simply invoke that method for each verticle appropriately

knotx-example/knotx-example-monolith/src/main/resources/knotx-example-monolith.json
```diff
@@ -12,7 +12,14 @@
         "repository.address": "template-repository",
         "engine.address": "template-engine"
       }
-    }
+    },
+    "worker" : false,
```
marcinczeczko: remove all scaling properties from example-monotlith fro all verticles - it does not make sense here

knotx-example/knotx-example-monolith/src/test/java/com/cognifide/knotx/monolith/ApplicationTestHelper.java
```diff
@@ -45,16 +45,16 @@ public static void startKnotx() throws Exception {
 
     knotxPort = configuration.getJsonObject("server").getJsonObject("config").getInteger("http.port");
 
-    vertx.deployVerticle(MockRemoteRepositoryVerticle.class.getName(), new DeploymentOptions().setConfig(
-        configuration.getJsonObject("mockRepo")));
+    vertx.deployVerticle(MockRemoteRepositoryVerticle.class.getName(),
```
marcinczeczko: You can make same helper method for verticle deployment as in Standalon - will be less code here.

knotx-example/knotx-mocks/src/main/resources/knotx-mocks.json
```diff
@@ -1,10 +1,28 @@
 {
   "mockRepo": {
-    "mock.data.root": "mock/repository",
-    "http.port": 3001
+    "config": {
+      "mock.data.root": "mock/repository",
+      "http.port": 3001
+    },
+    "worker" : false,
```
marcinczeczko: Remove all deployment options from mocks - does not make sense here

documentation/src/main/wiki/GettingStarted.md
```diff
@@ -260,3 +264,32 @@ This section configures the Services mock used by the example application. It co
...
+To deploy verticle with advanced options use following properties:
+```json
+{
+    ... 
+  "mockService": {
+    "config": {
+      "mock.data.root": "mock/service",
+      "http.port": 3000
+    },
+    "worker" : false,
```
Skejven: I think that it would be more natural to put all deployment options into `deploymentOptions` property:
```json
{
    ... 
    "mockService": {
        "mock.data.root": "mock/service",
        "http.port": 3000
    },   
    "deploymentOptions": {
      "instances": 2
    }
    ...
}
```

documentation/src/main/wiki/GettingStarted.md
```diff
@@ -260,3 +264,32 @@ This section configures the Services mock used by the example application. It co
...
+    "multiThreaded": false,
+    "isolationGroup": "null",
+    "ha": false,
+    "extraClasspath": [],
+    "instances": 2,
+    "isolatedClasses": []
```
Skejven: Do we need all other properties form deployment options defined? Moste of them are default values. I think it would be better to give user a reference to available [vert.x deployment options](http://vertx.io/docs/apidocs/io/vertx/core/DeploymentOptions.html).

knotx-core/knotx-standalone/src/main/resources/standalone.json
```diff
@@ -12,7 +12,14 @@
         "repository.address": "template-repository",
         "engine.address": "template-engine"
       }
-    }
+    },
```
Skejven: See my comment to json configuration.

documentation/src/main/wiki/GettingStarted.md
marcinczeczko: Not exacly - having it as is it now - you don't need to do anything. vertx expects in json deployment options on current level, while config object is automaticlly read also. if you put deployment options into "deploymentOptions" then you'd need to handle it manually

documentation/src/main/wiki/GettingStarted.md
marcinczeczko: I put same comment :)

knotx-core/knotx-standalone/src/main/resources/standalone.json
marcinczeczko: and see my previous comments

documentation/src/main/wiki/ProductionUsage.md
```diff
@@ -74,11 +75,12 @@ The *core* module contains 3 Knot.x verticle without any sample data. Here's how
           "port": 8080
         }
       ]
-    }
+    },
   }
 }
 ```
 Configuration JSON contains three config sections, one for each Knot.x verticle.
+Each verticle can be configured with additional [Deployment Options](https://github.com/Cognifide/knotx/wiki/GettingStarted#deployment-options)
```
Skejven: We can get the same link with `[[Deployment Options|GettingStarted#deployment-options]]` .

PR #95, opened 2016-10-10
Ensure mock repository server will return HTML always with UNIX newlines

marcinczeczko:


knotx-example/knotx-mocks/src/main/java/com/cognifide/knotx/mocks/service/MockRemoteRepositoryHandler.java
```diff
@@ -31,6 +31,7 @@
 public class MockRemoteRepositoryHandler implements Handler<HttpServerRequest> {
 
   private static final String SEPARATOR = "/";
+  final String DOS_NEWLINE = "\r\n", NIX_NEWLINE = "\n", MAC_NEWLINE = "\r";
```
Skejven: `private static final`

knotx-example/knotx-mocks/src/main/java/com/cognifide/knotx/mocks/service/MockRemoteRepositoryHandler.java
```diff
@@ -70,4 +71,8 @@ private String getContentPath(String path) {
     }
   }
 
+  private String normalizeEndline(String content) {
+    return content.replace(DOS_NEWLINE, NIX_NEWLINE).replace(MAC_NEWLINE, NIX_NEWLINE);
```
Skejven: Could we just get rid of all endlines (move all code to a single line) and end it with UNIX eol?

PR #97, opened 2016-10-10
Fix for line endings

wiiitek:
Tested on windows
with `git config --global core.autocrlf false`
after cloning a fresh knotx repository with `git status` command.


documentation/pom.xml
```diff
@@ -1,105 +1,105 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
wiiitek: line-endings only changed, add query strin of `?w=1` to github URL

wiiitek:
--- 

tomaszmichalak:
Approved.

PR #98, opened 2016-10-10
redundant method removed

wiiitek:
---


wiiitek:
there were two methods, that did the same thing:
- `JsonObject readConfig(String path)`
- `JsonObject readJson(String path)`

the *config* in method name is not needed.

PR #102, opened 2016-10-15
Bugfix/mock services fixes

marcinczeczko:
---


knotx-core/knotx-repository/knotx-repository-filesystem/src/main/java/com/cognifide/knotx/repository/FilesystemRepositoryVerticle.java
```diff
@@ -81,11 +82,15 @@ public void start() throws Exception {
...
     return fileSystem.openObservable(localFilePath, new OpenOptions())
         .flatMap(this::processFile)
-        .map(buffer -> new HttpResponseWrapper().setStatusCode(HttpResponseStatus.OK).setBody(buffer))
+        .map(buffer -> new HttpResponseWrapper().setStatusCode(HttpResponseStatus.OK).setHeaders(headers()).setBody(buffer))
         .defaultIfEmpty(new HttpResponseWrapper().setStatusCode(HttpResponseStatus.NOT_FOUND))
         .onErrorReturn(this::processError);
   }
 
+  private MultiMap headers() {
+    return MultiMap.caseInsensitiveMultiMap().add("Content-Type", "text/html; charset=utf-8");
```
wiiitek: Are we setting `Content-Type` header always to the same value?
Filesystem repository could probably provide images, CSS styles, JS fles.

I don't think Knot.X should be responsible for determining content type.
For HTTP repository this header is rewritten.
Could we live **wthout** `Content-Type` header at all if templates are retrieved from disk?

PR #103, opened 2016-10-16
Filesystem Repository should NOT create new files

wiiitek:
JAVADOC for `openObservable` says:

> The file is opened for both reading and writing. If the file does not already exist it will be created.

this needed to be fixed.


knotx-core/knotx-repository/knotx-repository-filesystem/src/main/java/com/cognifide/knotx/repository/FilesystemRepositoryVerticle.java
```diff
@@ -84,7 +84,7 @@ public void start() throws Exception {
 
     LOGGER.trace("Fetching file `{}` from local repository.", localFilePath);
 
-    return fileSystem.openObservable(localFilePath, new OpenOptions())
+    return fileSystem.openObservable(localFilePath, new OpenOptions().setCreate(false))
```
marcinczeczko: setWrite(false) is also required as it's true by default.
Instead of specifing it inline, please create final field of the class with OpenOptions set to write=false and create=false.

Skejven: Fixed.

PR #116, opened 2016-10-20
JsonAdapter interface and KnotMocks update

Skejven:
JsonAdapter interface
---


Skejven:
Declined.

PR #120, opened 2016-10-21
TemplateEngine tests

Skejven:
TemplateEngine tests - work in progress
---


knotx-core/knotx-engine-view/src/test/java/com/cognifide/knotx/engine/view/impl/TemplateEngineTest.java
```diff
@@ -0,0 +1,106 @@
...
+    // when
+    Observable<String> result = templateEngine.process(getContext("one-raw-fragment", 1));
+
+    // then
+    result.subscribe(
+        response -> {
+          final String expectedClean = Jsoup.clean(expectedResponse, Whitelist.relaxed());
+          final String responseClean = Jsoup.clean(expectedResponse, Whitelist.relaxed());
+          context.assertEquals(expectedClean, responseClean);
+          verify(eventBus, times(0)).sendObservable(anyString(), anyObject());
```
marcinczeczko: put some comment here, why we're expecting no calls on event bus will be done ? Btw. are you sure there's going to be sendObservable() ? what if there is a .send() call on event bus ?, I think .send() is better as it's actually wrapped by sendObvserable()

knotx-core/knotx-engine-view/src/test/resources/test-no-fragments.html
```diff
@@ -1,14 +0,0 @@
-<html>
```
marcinczeczko: Why you removed those test htmls ? there were tests using this, isn't ? or it's moved to splitter ?

Skejven: Splitter is using them now.

PR #139, opened 2016-10-25
knotx-134 - rename view engine into knot view

Skejven:
Refactored ViewEngine into KnotView
---


documentation/src/main/wiki/GettingStarted.md
```diff
@@ -179,9 +179,9 @@ The config node consists of:
 #### 1.3. Engine section
 ```json
   ...
-  "com.cognifide.knotx.engine.view.ViewEngineVerticle": {
+  "com.cognifide.knotx.engine.view.ViewKnotVerticle": {
```
marcinczeczko: you forgot changing package name in here and other places

PR #162, opened 2016-11-03
Implemented Authorization knot

marcinczeczko:
Core Authorization Knot. That can be used for simple implementations of authorizations.
How to use it:
1. Deploy AuthorizationKnot in your Knot.x instance (by adding following example verticle entry)
```json
    "com.cognifide.knotx.knot.authorization.AuthorizationKnotVerticle": {
      "config": {
        "address": "knotx.knot.auth",
        "adapters": [
          {
            "name": "auth-handler",
            "address": "auth-handler",
            "allowed.request.headers": [
              "Content-Type",
              "X-*"
            ],
            "allowed.response.headers": [
              "X-*"
            ]
          }
        ]
      }
    }
```
2. AuthorizationKnot uses service adapters, so you can point (**address**) field your http service adapter, setting up what headers must be passed to the target service, and what response headers will be respected.
3. Modify routing in **Server** verticle by adding auth knot, as follows:
```json
          "GET": [
            {
              "path": "/content/.*",
              "address": "knotx.knot.auth",
              "onTransition": {
                "next": {
                  "address": "knotx.knot.action",
                  "onTransition": {
                    "next" : {
                      "address": "knotx.knot.view"
                    }
                  }
                }
              }
            }
          ],
```
This will cause that each request will always pass auth Knot

4. On the HTML pages what need to be authorized, e.g. second step of login you have to add **auth** snippet in any place of that HTML (e.g. at the very begining). Keep in mind that having **auth** snippet on any page with **AuthorizationKnot** - that snippet will always be removed, so it must not have any content inside.
```html
<script data-api-type="auth"
        data-knotx-auth="auth-handler"
        data-knotx-on-unauthorized="/content/local/login/step1.html"
        type="text/x-handlebars-template">
</script>
<html>
  <h1>your content, or other snippets</h1>
</html>
```

5. Snippet parameters:

- **data-knotx-auth** - is the service adapter name available in your AuthorizationKnotVerticle configuration
- **data-knotx-on-unauthorized** - is the redirect path (307 - Temporary Redirect) if request is unauthorized

6. How it works

- If page does not have **auth** snippet -> AuthorizationKnot does not do any Authorization
- If page DOES HAVE **auth** snippet -> AuthorizationKnot sends request to configured Service Adapter
 -- If service Adapter respond with **200** status code -> It means access is authorized and page is processed normally
 -- If service Adapter response with other status code, e.g. **401** -> It means request is not authorized and 307 redirect will be returned to **data-knotx-on-unauthorized** path.


knotx-core/knotx-knot-authorization/pom.xml
```diff
@@ -0,0 +1,144 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Knot.x - Reactive microservice assembler - Action Knot Verticle
```
Skejven: Should be `Auhtorization Knot Verticle`. Please check other license headers in this module.

PR #165, opened 2016-11-04
Bugfix/placeholders are not encoded

marcinczeczko:
value encoded using RFC standard with the exceptions:
space will be %20 (instead of +)
slash (/) will remain the same - as it's not in standard but widely used and supported

### DIFF
--- a/knotx-adapters/knotx-adapter-common/src/main/java/com/cognifide/knotx/adapter/common/placeholders/UriTransformer.java
+++ b/knotx-adapters/knotx-adapter-common/src/main/java/com/cognifide/knotx/adapter/common/placeholders/UriTransformer.java
@@ -28,7 +28,12 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import io.vertx.core.logging.Logger;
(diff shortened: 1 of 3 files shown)


knotx-adapters/knotx-adapter-common/src/main/java/com/cognifide/knotx/adapter/common/placeholders/UriTransformer.java
```diff
@@ -60,4 +63,12 @@ private static String getPlaceholderValue(ClientRequest request, String placehol
         .findFirst()
         .orElse("");
   }
+
+  private static String encodeValue(String value) {
+    try {
+      return URLEncoder.encode(value, "UTF-8").replace("+", "%20").replace("%2F", "/");
+    } catch (UnsupportedEncodingException var3) {
+      throw new UnsupportedCharsetException("UTF-8");
```
Skejven: Maybe we should at least log `var3` ? (btw. please rename it to some standard `e`).

marcinczeczko: fixed

Issue #208, opened 2017-01-04
Update default logging configuration


Skejven:
By default, configured logging should log only at `INFO` level.
Current configuration in `logback.xml` files in the whole repository log at `TRACE` level.
Please refactor all `logback.xml` files and set logging to `INFO` level.

toniedzwiedz:
@Skejven do we also want to update the `logback-test.xml` files in a similar manner?

PR #224, opened 2017-01-14
#208 Changed log level from TRACE to INFO in all logback.xml files

toniedzwiedz:
Updated `logback.xml` files to use the `INFO` log level as opposed to `TRACE`, as described in #208 

Not sure about the `logback-test.xml` files. Do we want those updated too?

---


tomaszmichalak:
> Not sure about the logback-test.xml files. Do we want those updated too?

yes

PR #238, opened 2017-01-30
Fixed #237 - Do not call Fragment assembler if Knot processing result…

### LINKED ISSUE #237: Form submission failed
Action Knot sets response status code to 301. It means that user should be redirected to new location - we should not futher process this request. Currently Assebler processes request and responds with error.

marcinczeczko:
Fixed #237 - Do not call Fragment assembler if Knot processing finished with response Code other than 200
- Added unit test to cover this bug

### DIFF
--- a/knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
+++ b/knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
@@ -51,28 +51,34 @@ public class KnotxAssemblerHandler implements Handler<RoutingContext> {
   public void handle(RoutingContext context) {
     KnotContext knotContext = context.get("knotContext");
 
-    assembler.processObservable(knotContext)
-        .doOnNext(this::traceMessage)
-        .subscribe(
-            ctx -> {
-              if (ctx.getClientResponse().getStatusCode() == HttpResponseStatus.OK.code()) {
-                sendResponse(context, ctx);
-              } else {
-                context.fail(ctx.getClientResponse().getStatusCode());
+    if (knotContext.getClientResponse().getStatusCode() != HttpResponseStatus.OK.code()) {
+      HttpServerResponse httpResponse = context.response();
+      writeHeaders(httpResponse, knotContext.getClientResponse());
+      httpResponse.setStatusCode(knotContext.getClientResponse().getStatusCode()).end();
+    } else {
+      assembler.processObservable(knotContext)
+          .doOnNext(this::traceMessage)
+          .subscribe(
+              ctx -> {
+                if (ctx.getClientResponse().getStatusCode() == HttpResponseStatus.OK.code()) {
+                  sendResponse(context, ctx);
+                } else {
+                  context.fail(ctx.getClientResponse().getStatusCode());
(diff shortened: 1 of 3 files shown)


knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
```diff
@@ -51,28 +51,34 @@ static KnotxAssemblerHandler create(Vertx vertx, KnotxServerConfiguration config
...
-    assembler.processObservable(knotContext)
-        .doOnNext(this::traceMessage)
-        .subscribe(
-            ctx -> {
-              if (ctx.getClientResponse().getStatusCode() == HttpResponseStatus.OK.code()) {
-                sendResponse(context, ctx);
-              } else {
-                context.fail(ctx.getClientResponse().getStatusCode());
+    if (knotContext.getClientResponse().getStatusCode() != HttpResponseStatus.OK.code()) {
+      HttpServerResponse httpResponse = context.response();
```
tomaszmichalak: Why not sendResponse() method here? We can slightly modify this method to support both cases.

knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
```diff
@@ -87,9 +93,7 @@ private void writeHeaders(final HttpServerResponse response,
                     .getAll(name)
                     .forEach(value -> response.putHeader(name, value))
         );
-  }
 
-  private void writeContentLength(final HttpServerResponse response) {
```
tomaszmichalak: Nice, this method was not needed.

knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
```diff
@@ -51,28 +51,34 @@ static KnotxAssemblerHandler create(Vertx vertx, KnotxServerConfiguration config
...
 
-    assembler.processObservable(knotContext)
-        .doOnNext(this::traceMessage)
-        .subscribe(
-            ctx -> {
-              if (ctx.getClientResponse().getStatusCode() == HttpResponseStatus.OK.code()) {
-                sendResponse(context, ctx);
-              } else {
-                context.fail(ctx.getClientResponse().getStatusCode());
+    if (knotContext.getClientResponse().getStatusCode() != HttpResponseStatus.OK.code()) {
```
tomaszmichalak: Can we extract this condition to method?

marcinczeczko: fixed

knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
marcinczeczko: fixed

knotx-server/src/main/java/io/knotx/server/KnotxAssemblerHandler.java
marcinczeczko: :)

PR #243, opened 2017-01-31
Make action knot rewrite 'Content-Type' and 'Content-Length' headers

pawelkoper:


marcinczeczko:
Could you please check if there are more occurrences of this wrong string?

tomaszmichalak:
Checked.

PR #244, opened 2017-02-01
Feature/fixed url encoding post body

marcinczeczko:
---

### DIFF
--- a/knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
+++ b/knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
@@ -20,7 +20,7 @@ package io.knotx.adapter.common.http;
 import io.knotx.adapter.common.exception.AdapterServiceContractException;
 import io.knotx.adapter.common.exception.UnsupportedServiceException;
 import io.knotx.adapter.common.placeholders.UriTransformer;
-import io.knotx.adapter.common.post.FormBodyBuilder;
+import io.knotx.adapter.common.post.UrlEncodedBodyBuilder;
 import io.knotx.dataobjects.AdapterRequest;
 import io.knotx.dataobjects.ClientRequest;
 import io.knotx.dataobjects.ClientResponse;
@@ -143,7 +143,7 @@ public class HttpClientFacade {
       filteredHeaders.names().forEach(
           headerName -> httpRequest.putHeader(headerName, filteredHeaders.get(headerName)));
       if (!serviceRequest.getFormAttributes().isEmpty()) {
-        httpRequest.end(FormBodyBuilder.encodeBody(serviceRequest.getFormAttributes()));
+        httpRequest.end(UrlEncodedBodyBuilder.encodeBody(serviceRequest.getFormAttributes()));
       } else {
         httpRequest.end();
       }
--- a/knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
+++ b/knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
@@ -143,7 +143,7 @@ public class HttpClientFacade {
       filteredHeaders.names().forEach(
           headerName -> httpRequest.putHeader(headerName, filteredHeaders.get(headerName)));
       if (!serviceRequest.getFormAttributes().isEmpty()) {
-        httpRequest.end(FormBodyBuilder.createBody(serviceRequest.getFormAttributes()));
+        httpRequest.end(FormBodyBuilder.encodeBody(serviceRequest.getFormAttributes()));
       } else {
         httpRequest.end();
       }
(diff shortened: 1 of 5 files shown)


knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -143,7 +143,7 @@ protected ClientRequest buildServiceRequest(ClientRequest originalRequest, JsonO
       filteredHeaders.names().forEach(
           headerName -> httpRequest.putHeader(headerName, filteredHeaders.get(headerName)));
       if (!serviceRequest.getFormAttributes().isEmpty()) {
-        httpRequest.end(FormBodyBuilder.createBody(serviceRequest.getFormAttributes()));
+        httpRequest.end(UrlEncodedBodyBuilder.encodeBody(serviceRequest.getFormAttributes()));
```
tomaszmichalak: We should check 'Content-Type' header and based on this value decide how to build the body. Default should be 'application/x-www-form-urlencoded'.

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/post/UrlEncodedBodyBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+          sb.append('&');
+        }
+      }
+      return sb.toString();
+    }
+  }
+
+  private static String encodeComponent(String s, Charset charset) {
+    try {
+      return URLEncoder.encode(s, charset.name()).replace("+", "%20");
```
pawelkoper: I think this is wrong. URLEncoder will encode spaces to '+' sign and then you will replace it to "%20". So spaces and '+' character will be encoded to "%20". And it shuold be like that:
" " -> "+"
"+" -> "%20"

tomaszmichalak: This is reference implementation: io.netty.handler.codec.http.QueryStringEncoder.

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/post/UrlEncodedBodyBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+
+  public static String encodeBody(final MultiMap formAttributes, Charset charset) {
+    if (formAttributes == null || formAttributes.isEmpty()) {
+      return StringUtils.EMPTY;
+    } else {
+      Iterator<Entry<String, String>> entryIt = ((io.vertx.core.MultiMap) formAttributes
+          .getDelegate())
+          .iterator();
+
+      StringBuilder sb = new StringBuilder();
```
toniedzwiedz: Can we make an educated guess as to the initial size of the buffer? It's probably going to be more than 16 characters anyway so memory allocation for the buffer will happen multiple times.

knotx-adapter/knotx-adapter-common/src/test/java/io/knotx/adapter/common/post/UrlEncodedBodyBuilderTest.java
```diff
@@ -24,37 +24,37 @@
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Test;
 
-public class FormBodyBuilderTest {
+public class UrlEncodedBodyBuilderTest {
```
toniedzwiedz:
```
<disturbingMentalImage>
Imagine this has to do with body building. 
</disturbingMentalImage>
```
:)

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/post/UrlEncodedBodyBuilder.java
tomaszmichalak: What size do you suggest? In my opinion it could be 128 :)

toniedzwiedz: Depends on the form really. How many attributes do we usually get here?

tomaszmichalak: Depends on the client. Most forms are simple and have only few fields. So 128 looks good.

toniedzwiedz:
How does this relate to handling multipart forms? Do we support those?

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/post/UrlEncodedBodyBuilder.java
toniedzwiedz: @tomaszmichalak I don't mind. 128 sounds a lot more plausible than 16.

PR #254, opened 2017-02-05
Sonarqube bugfixes

Skejven:
---

### DIFF
--- a/knotx-core/src/main/java/io/knotx/adapter/AbstractAdapterProxy.java
+++ b/knotx-core/src/main/java/io/knotx/adapter/AbstractAdapterProxy.java
@@ -47,7 +47,13 @@ public abstract class AbstractAdapterProxy implements AdapterProxy {
         );
   }
 
-  private AdapterResponse getErrorResponse(Throwable error) {
+  /**
+   * Method generates error {@link AdapterResponse} in case of processing failure.
+   *
+   * @param error - error that occurred.
+   * @return - error response (e.g. with 500 status code and other info).
+   */
+  protected AdapterResponse getErrorResponse(Throwable error) {
     return new AdapterResponse().setResponse(new ClientResponse()
         .setStatusCode(HttpResponseStatus.INTERNAL_SERVER_ERROR.code())
         .setBody(Buffer.buffer(error.getMessage())));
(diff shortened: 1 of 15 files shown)


knotx-junit/src/main/java/io/knotx/junit/rule/TestVertxDeployer.java
```diff
@@ -60,9 +61,9 @@ public void evaluate() throws Throwable {
               }
               latch.countDown();
             });
-        latch.await(3, TimeUnit.SECONDS);
+        boolean await = latch.await(3, TimeUnit.SECONDS);
 
-        if (latch.getCount() != 0) {
+        if (!await || latch.getCount() != 0) {
```
toniedzwiedz: Isn't the second half of this condition superfluous?

In line 64:

    boolean await = latch.await(3, TimeUnit.SECONDS);

The `await` variable will have the value of `true` if the count has reached 0, `false` if the time elapsed but the count is not 0. Otherwise (if the thread is interrupted), it will throw an exception and the condition in the next line will not be reached.

Either way, I don't see how `(!await || latch.getCount() != 0)` could yield a different result than `!await`

knotx-mocks/src/main/java/io/knotx/mocks/adapter/MockRemoteRepositoryHandler.java
```diff
@@ -39,8 +39,7 @@
   private final JsonObject delayPerPath;
   private Set<String> textFileExtensions = Sets
```
toniedzwiedz: This probably could be made `final` or `static final`

knotx-junit/src/main/java/io/knotx/junit/rule/TestVertxDeployer.java
Skejven: Right, fixed.

knotx-mocks/src/main/java/io/knotx/mocks/adapter/MockRemoteRepositoryHandler.java
Skejven: Fixed.

knotx-core/src/main/java/io/knotx/adapter/AbstractAdapterProxy.java
```diff
@@ -42,12 +42,12 @@ public void process(AdapterRequest request, Handler<AsyncResult<AdapterResponse>
...
             error -> {
               LOGGER.error("Error happened during Adapter Request processing", error);
-              result.handle(Future.succeededFuture(getErrorResponse(request, error)));
+              result.handle(Future.succeededFuture(getErrorResponse(error)));
             }
         );
   }
 
-  protected AdapterResponse getErrorResponse(AdapterRequest request, Throwable error) {
+  private AdapterResponse getErrorResponse(Throwable error) {
```
Skejven: Fixed, added also a javadoc for this method since it may be now overridden.

PR #261, opened 2017-02-09
Feature/content extract improvement

tomaszmichalak:
---


knotx-core/pom.xml
```diff
@@ -100,6 +100,15 @@
       <groupId>uk.co.datumedge</groupId>
       <artifactId>hamcrest-json</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-all</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
```
wiiitek: Mockito has `provided` scope in *knotx-junit* module.
It looks like in core module the default scope could be used (test).
Or am I wrong? Could we remove thes scope override here?
(should have)

knotx-core/src/test/java/io/knotx/fragments/FragmentContentExtractorTest.java
```diff
@@ -37,8 +42,22 @@
   public void getUnwrappedContent_withFragment_expectDefinedContent(Fragment fragment,
```
toniedzwiedz: Looking at the test input, it seems we're only testing the happy paths here. How is the class under test supposed to behave on invalid input?

knotx-knot/knotx-knot-fragment-assembler/src/test/java/io/knotx/assembler/FragmentAssemblerTest.java
```diff
@@ -67,7 +72,7 @@ public void callAssemblerWithNoSnippets_expectInternalServerError(TestContext co
   @KnotxConfiguration("test.unwrap.io.knotx.FragmentAssembler.json")
   public void callAssemblerWithEmptySnippet_expectNoContentStatus(TestContext context)
       throws Exception {
-    callAssemblerWithAssertions(context, Collections.singletonList(" "),
+    callAssemblerWithAssertions(context, Collections.singletonList(new ImmutablePair<>(Collections.singletonList(RAW), " ")),
```
toniedzwiedz: What is the role of the space being passed as a string here?

knotx-core/src/test/java/io/knotx/fragments/FragmentContentExtractorTest.java
```diff
@@ -37,8 +42,22 @@
...
       String expectedContentFileName) throws Exception {
 
-    final String expectedContent = FileReader.readText(expectedContentFileName);
+    final String expectedContent = readText(expectedContentFileName);
     final String unwrappedContent = FragmentContentExtractor.getUnwrappedContent(fragment);
     assertThat(unwrappedContent, equalToIgnoringWhiteSpace(expectedContent));
   }
+
+  @Coercion
+  public Fragment provideFragment(String fragmentContentFile) throws IOException {
```
tomaszmichalak: The problem here is connected with circular dependency between knotx-core and  knotx-junit. Can we fix this issue in new PR?

knotx-knot/knotx-knot-fragment-assembler/src/test/java/io/knotx/assembler/FragmentAssemblerTest.java
tomaszmichalak: Blank string. Fixed to StringUtils.SPACE.

knotx-core/pom.xml
tomaszmichalak: Fixed.

knotx-knot/knotx-knot-fragment-assembler/src/test/java/io/knotx/assembler/FragmentAssemblerTest.java
toniedzwiedz: @tomaszmichalak TBH, `StringUtils.SPACE` is not at all better than `" "`

It's just as meaningless in this context. I'd rather introduce a constant describing its purpose. Is it a separator?

PR #268, opened 2017-02-21
DON'T MERGE - Examples of HTML tables replacing an image

toniedzwiedz:
Do not merge!!!
This is just a demonstration.


Skejven:
I think it is generally a good idea to move the data from the table in an image form to a table.
It is unfortunate, that the look of the table in a form of an image looks better than a markup table...

PR #270, opened 2017-02-24
Update KnotxServer default configuration to allow all *.html files be…

### LINKED ISSUE #269: Update KnotxServer default configuration to allow all *.html files
Current default configuration of `KnotxServer` supports only requests that request path match `/content/*` pattern.
Update default configuration, so that Knot.x with out-of-the-box configuration can support all html files (`*.html`).

Skejven:
Update KnotxServer default configuration to allow all (`*.`) GET requests.

Fixes #269 
---

### DIFF
--- a/knotx-example/knotx-example-app/src/test/resources/test.io.knotx.KnotxServer.json
+++ b/knotx-example/knotx-example-app/src/test/resources/test.io.knotx.KnotxServer.json
@@ -25,7 +25,7 @@
       "routing": {
         "GET": [
           {
-            "path": "/content/.*",
+            "path": ".*\\.html",
             "address": "knotx.knot.action",
             "onTransition": {
               "next": {
@@ -41,7 +41,7 @@
         ],
         "POST": [
           {
-            "path": "/content/.*",
+            "path": ".*\\.html",
             "address": "knotx.knot.action",
             "onTransition": {
               "next": {
--- a/knotx-server/src/main/resources/io.knotx.KnotxServer.json
+++ b/knotx-server/src/main/resources/io.knotx.KnotxServer.json
@@ -51,7 +51,7 @@
       "routing": {
         "GET": [
           {
-            "path": ".*\\.html",
+            "path": ".*",
             "address": "knotx.knot.service",
             "onTransition": {
               "next": {
--- a/knotx-server/src/main/resources/io.knotx.KnotxServer.json
+++ b/knotx-server/src/main/resources/io.knotx.KnotxServer.json
@@ -51,7 +51,7 @@
       "routing": {
         "GET": [
           {
-            "path": "/content/.*",
+            "path": ".*\\.html",
             "address": "knotx.knot.service",
             "onTransition": {
               "next": {
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![][travis img]][travis]
 [![][license img]][license]
+[![][central-repo img]][central-repo]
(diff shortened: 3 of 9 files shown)


knotx-example/knotx-example-app/src/test/resources/test.io.knotx.KnotxServer.json
```diff
@@ -25,7 +25,7 @@
       "routing": {
         "GET": [
           {
-            "path": "/content/.*",
+            "path": ".*\\.html",
```
marcinczeczko: I think example app we can leave as it is now. Routes /content

Skejven: Ok, how about the tests configs?

marcinczeczko: I think. For test it doesn't matter

marcinczeczko: Up to you 😊

toniedzwiedz: The way I understand it, the previous config would accept query strings and the new one doesn't. Is my reasoning correct? If so, is this the desired behaviour?

Skejven: Good eye @toniedzwiedz , however in this case it works differently.
This path is used in [KnotxServerVertice](https://github.com/Cognifide/knotx/blob/16c6479cc8f47d83c1628dc0414df4c31460e676/knotx-server/src/main/java/io/knotx/server/KnotxServerVerticle.java#L58) to provide proper routing. When you check [Route.pathRegex](http://vertx.io/docs/apidocs/io/vertx/ext/web/Route.html#pathRegex-java.lang.String-) method documentation it says:
```
Set the path prefix as a regular expression. If set then this route will only match request URI paths, the beginning of which match the regex. Only a single path or path regex can be set for a route.
```

So the part important here is only a request path, not whole request. And this works fine with this prefix :)

tomaszmichalak: I am not sure that we should limit default routing to `.*\\.html`. You can have some other URL policy which breaks this rule. Someone can set rewrite rules like this for example: [http://knotx.io/hello-rest-service/](http://knotx.io/hello-rest-service/). In my opinion last routing entry should be `.*`.

knotx-server/src/main/resources/io.knotx.KnotxServer.json
```diff
@@ -51,7 +51,7 @@
       "routing": {
         "GET": [
           {
-            "path": "/content/.*",
+            "path": ".*\\.html",
```
tomaszmichalak: I am not sure that we should limit default routing to `.*\\.html`. You can have some other URL policy which breaks this rule. Someone can set rewrite rules like this for example: [http://knotx.io/hello-rest-service/](http://knotx.io/hello-rest-service/). In my opinion last routing entry should be `.*`.

toniedzwiedz: Ah, so this is where the default I was looking for comes from :)

Skejven:
Updated default configuration, so that all GET requests are now processed.

PR #293, opened 2017-03-22
use vertx web client and single

Skejven:
I've replaced `HttpClient` with `WebClient` - please review this code thoroughly as the core functionality is refactored here.

---


knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -18,51 +18,46 @@
...
+    this.webClient = webClient;
     this.services = services;
   }
 
-  public Observable<ClientResponse> process(AdapterRequest message, HttpMethod method) {
-    return Observable.just(message)
-        .doOnNext(this::validateContract)
+  public Single<ClientResponse> process(AdapterRequest message, HttpMethod method) {
+    return Single.just(message)
+        .doOnSuccess(this::validateContract)
```
marcinczeczko: what will happen if validation fails, does the event will be lost ? before we had defaultIfEmpty to emit internal server error

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -124,28 +119,28 @@ protected ClientRequest buildServiceRequest(ClientRequest originalRequest, JsonO
...
-      }
-    });
+    final HttpRequest<Buffer> request = webClient
+        .request(method, serviceMetadata.getPort(), serviceMetadata.getDomain(),
+            serviceRequest.getPath());
+
+    MultiMap filteredHeaders = getFilteredHeaders(serviceRequest.getHeaders(),
+        serviceMetadata.getAllowedRequestHeaderPatterns());
+    filteredHeaders.names().forEach(
+        headerName -> request.putHeader(headerName, filteredHeaders.get(headerName)));
```
marcinczeczko: I think we should use filteredHeaders.getAll(headerName), as the multimap can have multiple items for the given key, otherwise we might lost e.g. multiple cookies if it's a cookie header.

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -154,11 +149,9 @@ private MultiMap getFilteredHeaders(MultiMap headers, List<Pattern> allowedHeade
         .collect(MultiMapCollector.toMultiMap(o -> o, headers::getAll));
   }
 
-  private Observable<ClientResponse> wrapResponse(HttpClientResponse response) {
-    return Observable.just(Buffer.buffer())
-        .mergeWith(response.toObservable())
-        .reduce(Buffer::appendBuffer)
-        .doOnNext(this::traceServiceCall)
+  private Single<ClientResponse> wrapResponse(HttpResponse<Buffer> response) {
+    return Single.just(response.body())
```
marcinczeczko: does it mean, the httpReposnse<Buffer> already have collected all fragments of buffers so no meed to reduce buffer as before ?

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
Skejven: If I understand correctly `defaultIfEmpty` was triggered only when `flatMap` in line `64` returned no events.
`validateContract` throws exception, sot whole stream emmits `onError` - this flow is still true with `Single`.
I can't see other way than `onSuccess` when we create a `Single` with `just` :)

According to emitting `defaultIfEmpty` after `wrapResponse`:
I'm wondering when it may happen that nothing is emitted on this level... Is that possible with some timeout/network error? Such events will rather emit `error` than nothing.
Do you see any case when `defaultIfEmpty` could be triggered?

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
Skejven: Thanks, fixed.

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
Skejven: Yes - class `HttpResponse` comes from new `vertx-web-client` and is should all the "buffering" job for us. We get full `body` here.

tomaszmichalak: Why not `response.bodyAsBuffer()` ?

tomaszmichalak: > Keep in mind that using this HttpResponse impose to fully buffer the response body and should be used for payload that can fit in memory.

Skejven: So what do you think - should we care about the case that service response exceed available memory? Is that even possible in reasonable solutions? How much could JSON response have, more than `1 MB`?

knotx-core/src/main/java/io/knotx/util/DataObjectsUtil.java
```diff
@@ -72,7 +72,7 @@ public static boolean equalsBody(io.vertx.core.buffer.Buffer self,
    * @return - hashcode of the given {@link MultiMap} object
    */
   public static int multiMapHash(MultiMap multiMap) {
-    io.vertx.core.MultiMap map = (io.vertx.core.MultiMap) multiMap.getDelegate();
+    io.vertx.core.MultiMap map = multiMap.getDelegate();
 
     return map.entries().stream().mapToInt(
```
tomaszmichalak: Maybe simply: multiMap.getDelegate().entries().stream()...

knotx-adapter/knotx-adapter-service-http/src/main/java/io/knotx/adapter/service/http/impl/HttpServiceAdapterProxyImpl.java
```diff
@@ -20,31 +20,31 @@
...
 
   private HttpClientFacade httpClientFacade;
 
   public HttpServiceAdapterProxyImpl(Vertx vertx, HttpAdapterConfiguration configuration) {
-    this.httpClientFacade = new HttpClientFacade(getHttpClient(vertx, configuration),
+    this.httpClientFacade = new HttpClientFacade(getWebClient(vertx, configuration),
         configuration.getServices());
   }
 
+  private WebClient getWebClient(Vertx vertx, HttpAdapterConfiguration configuration) {
```
tomaszmichalak: Please move this private method below protected one.

PR #294, opened 2017-03-22
Replace deprecated rx methods

Skejven:
Replaced deprecated Observable methods with its current replacements.

---


knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/impl/ServiceKnotProxyImpl.java
```diff
@@ -84,10 +84,7 @@ private KnotContext createSuccessResponse(KnotContext inputContext) {
   }
 
   private Observable<FragmentContext> compileHtmlFragment(Fragment fragment) {
```
tomaszmichalak: Why Observable as result?

knotx-mocks/src/main/java/io/knotx/mocks/knot/KnotContextKeys.java
```diff
@@ -80,9 +81,10 @@ String key() {
 
   Observable<Optional<Object>> mockValue(FileSystem fileSystem, String resourcePath) {
```
tomaszmichalak: I think we should check where observable is really required. I like Single.

knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/impl/ServiceKnotProxyImpl.java
Skejven: We should return an `Observable` here because latter processing concerns stream of events (processing all services, more than one is possible).

knotx-mocks/src/main/java/io/knotx/mocks/knot/KnotContextKeys.java
Skejven: We use `Observable` in `handle` method which is just more natural here because `Single` enables emitting only **one** message while in our case we support the situation with no messages (result is 404 status code). Using `Single` here is probably possible, but `Observable` is just more natural.

knotx-mocks/src/main/java/io/knotx/mocks/knot/KnotContextKeys.java
```diff
@@ -92,9 +94,10 @@ String key() {
         : buffer.toJsonArray());
   }
 
-  private Observable<Buffer> processFile(final AsyncFile asyncFile) {
+  private Single<Buffer> processFile(final AsyncFile asyncFile) {
     return Observable.just(Buffer.buffer())
```
tomaszmichalak: Can we exchange this merge/reduce with asyncFile.toObservable().toSingle()?

PR #296, opened 2017-03-23
Fixed #136 - Support for params on markup and config

### LINKED ISSUE #136: Http Service & Action Adapter - Extended support for params on markup & config
Http Service & Action Adapters should support additional parameters provided in params.
- Currently it's only "path"
- Should support "queryString" - to be added to the service call path
- Should support "headers" - optional headers to be send to service call - overwriting existing if any matches.

Skejven:
Support for additional query and header params in service calls.

---


documentation/src/main/wiki/HttpServiceAdapter.md
```diff
@@ -111,6 +119,19 @@ Example configuration of a [[Service Knot|ServiceKnot]]:
         "params": {
           "path": "/service/twitter/user/{header.userId}"
         }
+      },
+      {
+        "name" : "javabooks",
```
tomaszmichalak: Good example but can I display those books with any example page? If yes please provide here a link to this page with information how to run demo.

rkarwacki: An example is presented [on the blog](http://knotx.io/blog/hello-rest-service/). However, it only shows how to use a fixed query parameter. 
It would be beneficial to create a new demo page explaining how to use parametrized calls. I will extend the demo with such page.

documentation/src/main/wiki/HttpServiceAdapter.md
```diff
@@ -208,3 +229,64 @@ In response, external service returns:
...
+    ]
+  }
+```
+
+We can set query parameters sent to the service using the following snippet:
+
+```html
+<script data-knotx-knots="services,handlebars" type="text/knotx-snippet"
+    data-knotx-service="products"
+    data-knotx-params='{"queryParams":{"amount":"4"}}">
```
Skejven: It looks that something is wrong in this line. `'` opens the params, but it is closed with `"`.

PR #299, opened 2017-04-28
Feature/297 customize request routing outside knots

mariuszkubis:

Previously, every request in Knot.X was routed through: 
-Server
-RepositoryConnector
-FragmentSplitter
-customizable Knots (Engine)
-FragmentAssembler

This is now called a "default flow", a typical way of processing requests in Knot.X

Here, we introduce a new way of processing - "custom flow", which allows you to omit some blocks of the default flow. 

A custom flow consists of:
-Server
-Gateway knot which sets the knot context of the request
-customizable Knots (Engine)
-Response provider Knot, which returns the result to the client (like FragmentAssembler in default flow, but it doesn't operate on Fragments)

To test this solution, run Example App and visit http://localhost:8092/customFlow/remote/simple.json to retrieve a JSON message processed by a custom flow.


documentation/src/main/wiki/GatewayMode.md
```diff
@@ -0,0 +1,41 @@
+# Gateway Mode
+
+The gateway mode provides an alternative way of processing requests, alternative to than presented in [[Knot Routing section|KnotRouting]].
```
toniedzwiedz: > The gateway mode provides an alternative way of processing requests, alternative to than presented in [[Knot Routing section|KnotRouting]].

Some repetition and awkward grammar here. How about rewriting this as follows:

The gateway mode provides a way of processing requests alternative to that presented in the [[Knot Routing section|KnotRouting]].

documentation/src/main/wiki/GatewayMode.md
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+```
+
+Here, the gateway mode will work on paths beginning with `/customFlow`.
+First, a Knot, called a Gateway Knot, sets the Knot Context of the request. The rest of the routing configuration can be customized to your needs. 
+
+After the routing is over, the response is returned from a verticle called Response Provider. 
+
+//TODO an image of the routing simillar to this on Knot Routing page
```
toniedzwiedz: Will this be provided as part of this Pull Request?

documentation/src/main/wiki/GatewayMode.md
```diff
@@ -0,0 +1,41 @@
...
+
+Depending of your routing implementation, you can use the Gateway Mode to return the external services response
```
toniedzwiedz: Depending _on_

documentation/src/main/wiki/HighLevelArchitecture.md
```diff
@@ -13,7 +13,7 @@ Diagram below depicts high level Knot.x architecture.
 
 Custom business logic can be encapsulated in dedicated Knots / Adapters.
 
-Knot is module which defines custom step while [request routing](#KnotRouting). It can process custom
+Knot is module which defines custom step while [[request routing|KnotRouting]]. It can process custom
```
toniedzwiedz: Awkward grammar.

_A Knot is a module which defines a custom step in the process of [[request routing|KnotRouting]]._

documentation/src/main/wiki/KnotRouting.md
```diff
@@ -86,3 +86,6 @@ Knot performs its business logic and returns a transition. The transition define
 In some cases Knot can decide to break the route and redirect the user to a different page.
 
 When all Knots on the route processed the request or one of Knots break the routing, Server returns a response to the user.
+
+It is also possible to define custom request flow, skipping Repository Connector, Fragment Splitter and Fragment Assembler.
```
toniedzwiedz: _a_ custom request flow

documentation/src/main/wiki/Server.md
```diff
@@ -116,6 +121,7 @@ In short, by default, server does:
 - Listens on port 8092
 - Displays exception details on error pages (for development purposes)
 - Returns certain headers in Http Response to the client (as shown above)
+- Uses [[default Knot.X routing mechanism|KnotRouting]]
```
toniedzwiedz: Uses _the_ default Knot.x routing mechanism.

documentation/src/main/wiki/GatewayMode.md
Skejven: It should be done as part of this PR.

knotx-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
```diff
@@ -0,0 +1,71 @@
...
+import io.knotx.dataobjects.KnotContext;
+import io.knotx.knot.AbstractKnotProxy;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpHeaders;
+import rx.Single;
+
+public class RequestProcessorKnotProxyImpl extends AbstractKnotProxy {
+
+  private final static String RESPONSE = "{\"message\":\"This is a sample custom flow response\"}";
```
Skejven: Should this class be a part of gateway implementation? It rather looks like an example. Maybe in such case this should go to `knotx-examples` or even to wiki docs?

knotx-server/src/main/java/io/knotx/server/KnotxGatewayResponseProviderHandler.java
```diff
@@ -0,0 +1,115 @@
...
+          .subscribe(
+              ctx -> {
+                if (isOkClientResponse(ctx.getClientResponse())) {
+                  sendResponse(context, ctx.getClientResponse());
+                } else {
+                  context.fail(ctx.getClientResponse().getStatusCode());
+                }
+              },
+              error -> {
+                LOGGER.error("Error happened while communicating with {} engine", error,
```
Skejven: Missing one `{}` in this log message.

knotx-server/src/main/java/io/knotx/server/SupportedMethodsAndPathsHandler.java
```diff
@@ -38,24 +44,31 @@ public static SupportedMethodsAndPathsHandler create(KnotxServerConfiguration co
...
-          context.request().method());
-      context.fail(HttpResponseStatus.METHOD_NOT_ALLOWED.code());
-    } else if (shouldRejectPath) {
-      LOGGER.warn("Requested path {} is not supported based on configuration",
-          context.request().path());
+    String path = context.request().path();
+    HttpMethod method = context.request().method();
+    boolean pathNotAllowedInDefaultFlow = isPathNotPresentInFlow(path, defaultFlow);
+    boolean pathNotAllowedInCustomFlow = isPathNotPresentInFlow(path, customFlow);
+    if (pathNotAllowedInDefaultFlow && pathNotAllowedInCustomFlow) {
```
Skejven: This is a candidate for some unit tests.

knotx-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
rkarwacki: Absolutely, this class was here just to show some example use-case and should not be a part of the implementation. 
I've moved the `RequestProcessorKnot` to `knotx-example` module and updated the docs. I think having an example usage of Gateway Mode in the example application is beneficial.

documentation/src/main/wiki/GatewayMode.md
rkarwacki: Unfortunately, I do not have access to the images in editable form. 
If you can provide it, I will create the needed image.

knotx-server/src/main/java/io/knotx/server/SupportedMethodsAndPathsHandler.java
```diff
@@ -38,24 +44,31 @@ public static SupportedMethodsAndPathsHandler create(KnotxServerConfiguration co
...
+      LOGGER.warn("Requested path {} is not supported based on configuration", path);
       context.fail(HttpResponseStatus.NOT_FOUND.code());
+    } else if (!pathNotAllowedInDefaultFlow && isMethodDisallowedInFlow(method, defaultFlow)) {
```
Skejven: So.. what exactly should happen, when `defaultPath` and `defaultMethod` are not allowed, but `customPath` and `customMethod` are allowed :) ?
It looks that we finish without any action on context (`tip`: there is double `if`).

Skejven:
I've refactored a bit and (I think that I...) fixed `SupportedMethodsAndPathsHandler` case I wrote before.
Please check if this is still valid :)
I also added a unit test for this handler.

Please correct me if I'm wrong - this `custom flow` feature enables defining only a single custom routing? It is not possible to create two or more custom flows?

rkarwacki:
@Skejven - that looks great.

Yes, it is possible to define only one `custom flow`. Why? Because you can still process your request in any way you want by plugging your custom logic between `GatewayKnot` and `ResponseProviderKnot`. They provide a minimal functionality to process _any_ request - setting up the `RequestContext` and returning the response. We couldn't figure out a case when someone might need a few `custom flows`.

Similarly, you can only have one `KnotRouting` in the `default flow` - once the fragments are split, you can process them in any way you want, and the universal way of returning them is to route them through the `FragmentAssembler`. Having a second `KnotRouting` would not be beneficial because in every case, they could have been merged into one. The same applies to the `Gateway Mode` feature.

Skejven:
Thanks @rkarwacki for explaining. I've added smal code refactors and I think work is ready to merge. This is great feature, thank you guys!

Issue #305, opened 2017-06-08
Allow passing additional parameters to action adapter from template


rkarwacki:
Consider a scenario where I want to submit a form and perform some custom business logic, like encrypting the data in Knot.X before sending it to external service. 

I want to perform my custom logic in a dedicated `ActionAdapter`. However, I cannot pass additional parameters (which may be needed to perform the logic) to adapter from template. I'd like to be able to define my form like this:

```
<script data-knotx-knots="form-1,services,handlebars"type="text/knotx-snippet">
   <form data-knotx-action="my-action" 
             data-knotx-on-error="error.html" 
             data-knotx-on-success="success.html" 
             data-knotx-adapter-params='{"secretKey":"mySecretKey"}' method="post">
      <input type="text" title="Name" name="name">
      <input type="hidden" value="UTF-8" name="_charset_">
   </form>
</script>
```

Then I could access the `secretKey` property in `AdapterRequest` object in my custom `Adapter` and use it to perform my business logic.

PR #306, opened 2017-06-09
Feature/305 allow passing additional parameters to adapter from template

rkarwacki:
My suggested solution for issue #305.
 

knotx-core/src/main/java/io/knotx/dataobjects/AdapterRequest.java
```diff
@@ -27,6 +27,8 @@
 
   private JsonObject params;
 
+  private JsonObject additionalAdapterParams;
```
Skejven: How about renaming this field to simpler `adapterParams` or `actionParams`?

rkarwacki: Agreed. Renamed to `adapterParams`.

PR #308, opened 2017-07-10
rephrase some sentences to be more clear

wiiitek:


documentation/src/main/wiki/KnotxModules.md
```diff
@@ -94,4 +94,4 @@ E.g.: `src/main/resources/my.custom.Service.json`
   ]
 }
 ```
-4. Override default configuration if necessary directly in starter JSON, or through JVM properties.
+4. If necessary, override default configuration directly in starter JSON, or through JVM properties.
```
tomaszmichalak: the default ...

PR #316, opened 2017-07-26
HTTP request body passed to Gateway Knots.

tomaszmichalak:
---

### DIFF
--- a/knotx-example/knotx-example-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
+++ b/knotx-example/knotx-example-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
@@ -60,7 +60,7 @@ public class RequestProcessorKnotProxyImpl extends AbstractKnotProxy {
     String responseBody = getResponseBodyAsString(inputContext);
 
     clientResponse.setBody(Buffer.buffer(responseBody))
-        .setHeaders(getHeaders(clientResponse, responseBody));
+        .setHeaders(getHeaders(clientResponse, responseBody.length()));
     clientResponse.setStatusCode(HttpResponseStatus.OK.code());
 
     return new KnotContext()
@@ -68,10 +68,10 @@ public class RequestProcessorKnotProxyImpl extends AbstractKnotProxy {
         .setClientResponse(clientResponse);
   }
 
-  private MultiMap getHeaders(ClientResponse clientResponse, String response) {
+  private MultiMap getHeaders(ClientResponse clientResponse, int bodyLength) {
     MultiMap headers = clientResponse.getHeaders();
     headers.add(HttpHeaders.CONTENT_LENGTH.toString().toLowerCase(),
-        Integer.toString(response.length()))
+        Integer.toString(bodyLength))
         .add("Content-Type", "application/json");
     return headers;
   }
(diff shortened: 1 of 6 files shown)


knotx-example/knotx-example-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
```diff
@@ -53,19 +56,30 @@ protected KnotContext processError(KnotContext knotContext, Throwable error) {
...
+    clientResponse.setBody(Buffer.buffer(responseBody))
+        .setHeaders(getHeaders(clientResponse, responseBody));
+    clientResponse.setStatusCode(HttpResponseStatus.OK.code());
 
     return new KnotContext()
         .setClientRequest(inputContext.getClientRequest())
         .setClientResponse(clientResponse);
   }
+
+  private MultiMap getHeaders(ClientResponse clientResponse, String response) {
```
marcinczeczko: Looks like you don't need `response` parameter here

knotx-example/knotx-example-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
```diff
@@ -53,19 +56,30 @@ protected KnotContext processError(KnotContext knotContext, Throwable error) {
...
+    MultiMap headers = clientResponse.getHeaders();
+    headers.add(HttpHeaders.CONTENT_LENGTH.toString().toLowerCase(),
+        Integer.toString(response.length()))
+        .add("Content-Type", "application/json");
+    return headers;
+  }
+
+  private String getResponseBodyAsString(KnotContext inputContext) {
```
marcinczeczko: Looks like you need this responsnse body only in line 62. If yes, return in this private method a Buffer object directly, so line 62 will looks like below
```
clientResponse.setBody(getResponseBody(inputContext))
```

knotx-server/src/main/java/io/knotx/server/KnotxServerVerticle.java
```diff
@@ -78,13 +77,17 @@ public void start(Future<Void> fut) throws IOException, URISyntaxException {
       );
     });
 
-    if(configuration.getCustomFlow().getEngineRouting() != null) {
+    if (configuration.getCustomFlow().getEngineRouting() != null) {
       configuration.getCustomFlow().getEngineRouting().forEach((key, value) -> {
-        if (key == HttpMethod.POST) {
+        if (key != HttpMethod.GET) {
```
marcinczeczko: What is request method is *HEAD* ?

knotx-example/knotx-example-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
tomaszmichalak: Required to count content length.

knotx-server/src/main/java/io/knotx/server/KnotxServerVerticle.java
tomaszmichalak: Fixed. Only POST, PUT, DELETE allowed.

knotx-example/knotx-example-gateway/src/main/java/io/knotx/gateway/impl/RequestProcessorKnotProxyImpl.java
marcinczeczko: So it's more clear if the parameter was simply a body length. currently it suggests you're doing something with the body

tomaszmichalak:
All CR requests fixed.

PR #317, opened 2017-08-04
Processing correct error message to console.

tomaszmichalak:
```

---

```


marcinczeczko:
It's worth to step whole code for. getCause () to verify if we're not swallowing more

Issue #247, opened 2017-02-02
Support for form submission multipart encoding 


tomaszmichalak:
We support only:
* `application/x-www-form-urlencoded (the default)`
* `text/plain`

We need to add support for `multipart/form-data`. In current implementation we throw excetion for header `Content-Type="multipart/form"`.

PR #319, opened 2017-08-06
#247 added file support from files-upload directory

Shaihuludus:
Added files support. Now files from file-uploads directory can be processed in gateway knots


knotx-server/src/main/java/io/knotx/server/KnotxGatewayContextHandler.java
```diff
@@ -61,6 +66,12 @@ private void handleRoute(final RoutingContext context, final String address) {
     if (StringUtils.isNotBlank(bodyAsString)) {
       knotContext.setFragments(Collections.singletonList(Fragment.raw(bodyAsString)));
     }
+    Set<FileUpload> fileUploads = context.fileUploads();
+    if (fileUploads != null) {
+      knotContext.setFilesData(fileUploads.stream().map(fileUpload -> {
+        return new FileData(fileUpload);
+      }).collect(Collectors.toList()));
```
toniedzwiedz: What's the reason we're switching between `Set` and `List`? Is this a limitation of an existing API or is `List` simply more suitable for the `knotContext` domain?

knotx-core/src/main/java/io/knotx/dataobjects/KnotContext.java
```diff
@@ -91,6 +94,15 @@ public KnotContext setFragments(List<Fragment> fragments) {
     return this;
   }
 
+  public List<FileData> getFilesData() {
```
toniedzwiedz: I'm not familiar with the way this works with Vert.x data objects but we're binding an API to a concrete class here. Perhaps it would be more flexible to use an interface to define the contract of the generic param instead? Or am I overthinking this?

Shaihuludus: Vert.x keeps information about files in defined structured but can't transformed them into json which is required by knot.x, That is why new structure is created. And it is not API so we don't need generic

knotx-server/src/main/java/io/knotx/server/KnotxGatewayContextHandler.java
Shaihuludus: True, changed to Set

knotx-core/src/main/java/io/knotx/dataobjects/FileData.java
```diff
@@ -0,0 +1,175 @@
+/*
+ * Copyright (C) 2016 Cognifide Limited
```
Shaihuludus: Can be

Skejven: Actually, if I understand the idea correctly - this is an `inception year`. And 2016 as Knot.x inception year is ok.
In [this post](http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/) you can read a wider discussion about the topic.

knotx-server/src/main/java/io/knotx/server/KnotxGatewayContextHandler.java
```diff
@@ -61,6 +66,12 @@ private void handleRoute(final RoutingContext context, final String address) {
     if (StringUtils.isNotBlank(bodyAsString)) {
       knotContext.setFragments(Collections.singletonList(Fragment.raw(bodyAsString)));
     }
+    Set<FileUpload> fileUploads = context.fileUploads();
+    if (fileUploads != null) {
+      knotContext.setFilesData(fileUploads.stream().map(fileUpload -> {
```
Skejven: You may do it also like this:
`knotContext.setFilesData(fileUploads.stream().map(FileData::new).collect(Collectors.toSet()))`

knotx-core/src/main/java/io/knotx/dataobjects/FileData.java
```diff
@@ -0,0 +1,175 @@
+/*
+ * Copyright (C) 2017 Cognifide Limited
```
Skejven: Actually, if I understand the idea correctly - this is an `inception year`. And 2016 as Knot.x inception year is ok.
In [this post](http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/) you can read a wider discussion about the topic.

marcinczeczko:
Please also update wiki documentation about KnotxContext as you added new field: https://github.com/Cognifide/knotx/wiki/Knot#knot-context

marcinczeczko:
@Skejven, @tomaszmichalak - I refactored the PR as it's pretty usefull feature. Added also a configurable limit for uploads.

TODO:
- unit tests - add tests for fileuploads
- documentation update (knotxContext, etc.)

Skejven:
I believe that this feature isn't ready to be merged.
If I understand correctly, `FileData` contains only references to uploaded file and this file is stored **locally** in the *file-uploads* directory.
That means, if we send `FileData` inside `KnotContext` to another `Verticle` that is working on different instance it will have no access to the uploaded file.
On the other hand, packaging the uploaded files content into `FileData` is absolutely not an option, because moving such big messages through the Event Bus will be slow (and uploaded file will not be consumed by every Knot/Adapter).
In my opinion, we should give more thoughts to this feature, and propose some solution that will enable access to the file in the distributed environment (MongoDB / Redis?).
Also it should be possible to access uploaded files in Adapters (especially Action Adapter), currently, Adapters doesn't have access to `KnotContext` and this is intentional (e.g. Adapters shouldn't be able do perform operations on Fragments).

tomaszmichalak:
No action with this PR is expected. We need to change an approach for uploaded files.

PR #328, opened 2017-08-25
feature/#326

DariuszBlicharski:
If NPE is thrown it will be passed to rxjava.
If one module or more specified in config are not found warning will be
produced and kntox will start with the rest of the modules

### DIFF
--- a/knotx-core/src/main/java/io/knotx/launcher/KnotxStarterVerticle.java
+++ b/knotx-core/src/main/java/io/knotx/launcher/KnotxStarterVerticle.java
@@ -33,26 +33,26 @@ public class KnotxStarterVerticle extends AbstractVerticle {
   @Override
   public void start(Future<Void> startFuture) throws Exception {
     printLogo();
-    Observable.defer(() -> Observable.from(config().getJsonArray("modules"))
-      .flatMap(module -> deployVerticle(module)
-        .onErrorResumeNext(this::verticleCouldNotBeDeployed)
-      )
-      .compose(joinDeployments())
-    )
-    .subscribe(
-      message -> {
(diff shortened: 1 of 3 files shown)


tomaszmichalak:
Please update documentation and explain based on some example: specify what do we expect in logs.

knotx-core/src/main/java/io/knotx/launcher/KnotxStarterVerticle.java
```diff
@@ -33,19 +33,27 @@
...
-            message -> {
-              LOGGER.info("Knot.x STARTED {}", message);
-              startFuture.complete();
-            },
-            error -> {
-              LOGGER.error("Verticle could not be deployed {}", error);
-              startFuture.fail(error);
-            }
-        );
+    Observable.defer(() -> Observable.from(config().getJsonArray("modules"))
```
Skejven: I'm not sure `defer` is necessary here. We have only single subscriber in this situation, and created `Observable` is not exposed outside this method.

knotx-core/src/main/java/io/knotx/launcher/KnotxStarterVerticle.java
```diff
@@ -33,19 +33,27 @@
...
+      },
+      error -> {
+        LOGGER.error("Verticle could not be deployed", error);
+        startFuture.fail(error);
+      }
+    );
+  }
+
+  private Observable<Pair<String, String>> verticleCouldNotBeDeployed(Throwable throwable) {
+    LOGGER.error(throwable.getMessage());
```
Skejven: Maybe we could pass here also a `module` so that the information about the error will be more descriptive.

PR #341, opened 2017-10-19
Adding blocking adapter proxy skeleton.

tomaszmichalak:
---


knotx-core/src/main/java/io/knotx/adapter/BlockingAdapterProxy.java
```diff
@@ -0,0 +1,60 @@
+package io.knotx.adapter;
```
Skejven: Looks like license header is missing.

tomaszmichalak: Done.

Issue #344, opened 2017-11-07
Wrong HTTP method used on a supported path doesn't produce any log message


jwadolowski:
## Steps to reproduce

* deploy knot.x extension that handles `POST /services/search/products/details` requests
* make sure routing is configured accordingly (take a look at config snippet below)
* send `GET` request to `/services/search/products/details`

## Expected state

* 404 response & information in the logs about unsupported path

## Current state

* 404 gets back to the client
* Lack of any message in the logs

## Noteworthy factoids

* call to a path that's **not** listed in the configuration anywhere causes expected log message: `WARN  i.k.s.SupportedMethodsAndPathsHandler - Requested path /test is not supported based on configuration`
* log level was increased to `TRACE`, but that didn't help
* configuration snippet that may be useful

```javascript
"customFlow": {
    "...",
    "routing": {
        "GET": [
            "..."
        ],
        "POST": [
            {
                "address": "knotx.gateway.gatewayknot",
                "onTransition": {
                    "next": {
                        "address": "search.products.details"
                    }
                },
                "path": "/services/search/products/details"
            },
            "..."
        ]
    }
}
```

marcinczeczko:
I see that you're actually getting the WARN message in LOG about that situation, whilest in "current state" you pointed out that no appropriate logs.
Could you please clarify what's your expectations?

jwadolowski:
The problem occurs when you use `GET` method on a supported path (defined in routing) that accepts only `POST` requests, i.e `POST /my/endpoint` is supported, but I sent `GET /my/endpoint` instead.

Proper log message gets printed when you ask for a path that's not present in routing configuration, i.e. `/test` in my example above.

marcinczeczko:
Fixed in #347

PR #347, opened 2017-11-23
Fixed #337, #344 & improved logging error situations

### LINKED ISSUE #337: Custom headers in knot.x HTTP calls
As of right now it's not possible to determine whether request comes from knot.x or some other HTTP client (w/o extra work on different layers), as by default it transparently passes on all the headers from original request.

A few use cases:
* HTTP routing for repository requests (request source is taken into consideration)
* integration with external services (i.e. some API may require extra HTTP header for the sake of authentication and/or tracking purposes)

In general it'd be great to have an ability to add headers to any HTTP call that knot.x does. I'm also thinking about generic header attached to every request that contains metadata about the source (something similar to `User-Agent`).

marcinczeczko:
- better logging for Server routing, http service adapter, http respository connector
- added custom header response/request - to help easier create an infrastructure
- small polisher of other aspects :)

### DIFF
--- a/documentation/src/main/wiki/HttpRepositoryConnector.md
+++ b/documentation/src/main/wiki/HttpRepositoryConnector.md
@@ -69,7 +69,7 @@ Main options available.
 | `address`                   | `String`                            | &#10004;       | Event Bus address of Http Repository Connector Verticle |
 | `clientOptions`             | `HttpClientOptions`                 | &#10004;       | HTTP Client options used when communicating with the destination repository. See [HttpClientOptions](http://vertx.io/docs/apidocs/io/vertx/core/http/HttpClientOptions.html) to get all options supported.|
 | `clientDestination`         | `JsonObject`                        | &#10004;       | Allows to specify **domain** and **port** of the HTTP Repository endpoint |
-| `customRequestHeader`       | `JsonObject`                        |                | Allows to specify header name and its value. The header will be send in each request to the configured services. |
+| `customRequestHeader`       | `JsonObject`                        |                | Allows to specify header **name** and its **value**. The header will be send in each request to the configured services. |
 
 ### Destination options
 
--- a/documentation/src/main/wiki/HttpRepositoryConnector.md
+++ b/documentation/src/main/wiki/HttpRepositoryConnector.md
@@ -27,17 +27,23 @@ Default configuration shipped with the verticle as `io.knotx.HttpRepositoryConne
         "domain": "localhost",
         "port": 3001
       },
+      "customRequestHeader": {
+        "name": "Server-User-Agent",
+        "value": "Knot.x"
+      },
       "allowedRequestHeaders": [
         "Accept*",
         "Authorization",
         "Connection",
         "Cookie",
         "Date",
+        "Edge*",
         "Host",
         "If*",
         "Origin",
         "Pragma",
         "Proxy-Authorization",
+        "Surrogate*",
         "User-Agent",
         "Via",
         "X-*"
@@ -45,6 +51,7 @@ Default configuration shipped with the verticle as `io.knotx.HttpRepositoryConne
     }
   }
 }
+
 ```
 In general, it:
 - Listens on event bus address `knotx.core.repository.http` for requests to the repository
@@ -62,6 +69,7 @@ Main options available.
 | `address`                   | `String`                            | &#10004;       | Event Bus address of Http Repository Connector Verticle |
 | `clientOptions`             | `HttpClientOptions`                 | &#10004;       | HTTP Client options used when communicating with the destination repository. See [HttpClientOptions](http://vertx.io/docs/apidocs/io/vertx/core/http/HttpClientOptions.html) to get all options supported.|
 | `clientDestination`         | `JsonObject`                        | &#10004;       | Allows to specify **domain** and **port** of the HTTP Repository endpoint |
+| `customRequestHeader`       | `JsonObject`                        |                | Allows to specify header name and its value. The header will be send in each request to the configured services. |
 
 ### Destination options
 
--- a/knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpAdapterConfiguration.java
+++ b/knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpAdapterConfiguration.java
@@ -15,7 +15,9 @@
  */
 package io.knotx.adapter.common.http;
 
+import com.google.common.collect.Lists;
 import io.knotx.http.StringToPatternFunction;
+import io.vertx.core.http.HttpClientOptions;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import java.util.List;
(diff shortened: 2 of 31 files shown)


knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpAdapterConfiguration.java
```diff
@@ -15,20 +15,30 @@
...
+
   private String address;
 
   private List<ServiceMetadata> services;
 
   private JsonObject clientOptions;
 
+  public HttpAdapterConfiguration() {
+    customRequestHeader = new JsonObject();
+    services = Lists.newArrayList();
```
Skejven: Can we set `Collections.emptyList()` here?

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -49,23 +51,44 @@
...
-        .flatMap(serviceRequest -> callService(serviceRequest, method))
+        .flatMap(
+            serviceRequest -> callService(serviceRequest, method)
+                .doOnSuccess(resp -> logResponse(serviceRequest, resp)))
         .flatMap(this::wrapResponse);
   }
 
+  private void logResponse(Pair<ClientRequest, ServiceMetadata> request,
+      HttpResponse<Buffer> resp) {
+    if (resp.statusCode() != HttpResponseStatus.OK.code()) {
```
Skejven: Not sure if we should `WARN` about other status codes than `200`.
Statuses other than `200` are sometimes used to communicate status of REST operation (e.g. like `201`, `203` or `204`).
Maybe this should be `DEBUG` level here or `WARN` only `4XX` and `5XX` responses and `TRACE` others?

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -77,17 +100,12 @@ protected void validateContract(AdapterRequest message) {
...
-   * <br>
-   * The responsibility of the method is to build ClientRequest based on the original Http
-   * Request<br>
-   * - It must set path property of the request based on the params<br>
-   * - It might set headers of the request if needed.<br>
-   * <br>
-   * In case of headers created modified in this method, ensure that your service configuration
-   * allows passing those headers to the target service. See 'allowedRequestHeaders' section
-   * of the configuration <br>
+   * Method responsible for building request to the service. <br> <br> The responsibility of the
```
Skejven: Can you please bring back formatting ?

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
tomaszmichalak: Agree.

knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/service/ServiceEngine.java
```diff
@@ -83,7 +85,9 @@ private JsonObject buildResultObject(AdapterResponse adapterResponse) {
     } else if (rawData.charAt(0) == '{') {
       object.put(RESULT_NAMESPACE_KEY, new JsonObject(rawData));
     } else {
-      throw new DecodeException("Result is neither Json Array nor Json Object");
+      LOGGER.error("Result of [{} {}] neither Json Array nor Json Object: [{}]",
```
tomaszmichalak: Should we hide this exception here? It changes a service error scenario behaviour.

knotx-repository-connector/knotx-repository-connector-http/src/main/java/io/knotx/repository/impl/RepositoryConnectorProxyImpl.java
```diff
@@ -143,22 +146,44 @@ private String encodeParamValue(String value) {
...
 
-  private ClientResponse toSuccessResponse(Buffer buffer,
-      final HttpClientResponse httpClientResponse) {
-    return new ClientResponse()
-        .setStatusCode(httpClientResponse.statusCode())
-        .setHeaders(httpClientResponse.headers())
-        .setBody((io.vertx.core.buffer.Buffer) buffer.getDelegate());
+  private ClientResponse toResponse(Buffer buffer, final HttpClientResponse httpResponse) {
+    if (httpResponse.statusCode() >= 300 && httpResponse.statusCode() < 400) { //redirect responses
+      LOGGER.warn("Repository 3xx response: {}, Headers[{}]", httpResponse.statusCode(),
```
tomaszmichalak: info?

documentation/src/main/wiki/HttpRepositoryConnector.md
```diff
@@ -27,24 +27,31 @@ Default configuration shipped with the verticle as `io.knotx.HttpRepositoryConne
         "domain": "localhost",
         "port": 3001
       },
+      "customRequestHeader": {
```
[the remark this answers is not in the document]
Skejven: Actually, we had discussion about it with @marcinczeczko . The reason here why it is just a single item, is that this mechanism should be just a marker about the origin of a request. All necessary logic connected with headers manipulation should be done at the different layer (e.g. Apache/Ngnix).

knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/service/ServiceEngine.java
marcinczeczko: We're not hiding anything here ? Before it was the exception generated here that broke the whole processing. But in fact instead it should log the error that data is not parsable and return empty json.

documentation/src/main/wiki/HttpRepositoryConnector.md
```diff
@@ -27,24 +27,31 @@ Default configuration shipped with the verticle as `io.knotx.HttpRepositoryConne
...
+        "name": "Server-User-Agent",
+        "value": "Knot.x"
+      },
       "allowedRequestHeaders": [
```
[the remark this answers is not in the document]
marcinczeczko: Now we shouldn't. If it's defined it's always to be send.

knotx-adapter/knotx-adapter-common/src/main/java/io/knotx/adapter/common/http/HttpClientFacade.java
```diff
@@ -49,19 +50,40 @@
...
-        .flatMap(serviceRequest -> callService(serviceRequest, method))
+        .flatMap(
+            serviceRequest -> callService(serviceRequest, method)
+                .doOnSuccess(resp -> logResponse(serviceRequest, resp)))
         .flatMap(this::wrapResponse);
   }
 
+  private void logResponse(Pair<ClientRequest, ServiceMetadata> request,
+                           HttpResponse<Buffer> resp) {
+    if (resp.statusCode() >= 400 && resp.statusCode() < 600) {
```
Skejven: Do we want to `trace` other status codes?

marcinczeczko: Good point. I'll check and will add trace for it and then do the merge

PR #348, opened 2017-11-23
Additional helpers

magdu:
Additional helpers for knotx templates
---


marcinczeczko:
Thanks for you contribution. It great stuff.
However, it seems that it's a perfect fit for Knot.x core helpers. Your PR just add it as example one into example application.
So, I suggest that instead you should put your helpers here: https://github.com/Cognifide/knotx/tree/master/knotx-knot/knotx-knot-handlebars/src/main/java/io/knotx/knot/templating/helpers

So they will become a core Knot.x helpers.

Skejven:
Hi @magdu , thank you for your contribution, the work is fantastic.
However, I disagree with @marcinczeczko and those helpers should not be a part of Knot.x core. Let's discuss it offline.
Thanks!

Skejven:
Please move this Pull Request to https://github.com/Knotx/knotx-handlebars-extension

Thank you!

PR #349, opened 2017-11-28
Enabling event bus delivery options for communication between Server …

tomaszmichalak:
---


knotx-core/src/main/java/io/knotx/proxy/reactive/AdapterProxyFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import io.vertx.reactivex.core.Vertx;
+import java.util.Optional;
+
+public final class AdapterProxyFactory {
+
+  private AdapterProxyFactory() {
+    // hidden
+  }
+
+  public static AdapterProxy createProxy(Vertx vertx, Optional<DeliveryOptions> deliveryOptions, String address) {
```
toniedzwiedz: Not sure how this fits into the architecture of this area of Knot.x but consuming an Optional as a parameter seems a bit strange. How about overloading the method to either accept the `DeliveryOptions` or not?

    public static AdapterProxy createProxy(Vertx vertx, String address, DeliveryOptions deliveryOptions)
    //...
    public static AdapterProxy createProxy(Vertx vertx, String address)

The caller either has the delivery options or not. If they don't, with the current API, they need to pass an empty optional here.

I can see this pattern in more places and I'm wondering what the benefits of using an `Optional` this way are.

knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java
```diff
@@ -31,5 +32,9 @@ static AdapterProxy createProxy(Vertx vertx, String address) {
     return new AdapterProxyVertxEBProxy(vertx, address);
   }
 
+  static AdapterProxy createProxyWithOptions(Vertx vertx, String address, DeliveryOptions deliveryOptions) {
```
marcinczeczko: why you added the new createProxy method ? can it be just one createProxy() method consuming deliveryOptions ?
I see that you created Factory specifically for instantiation of Knot's, Adapters, etc. that uses createProxy method with or without deliveryOptions. But it's not needed at all. A ....VertxEBProxy implementation handles it for you - if you will set deliveryOptions to null it will not use it - you don't need to do any handling or factories.

knotx-core/src/main/java/io/knotx/proxy/KnotProxy.java
```diff
@@ -30,5 +31,9 @@ static KnotProxy createProxy(Vertx vertx, String address) {
     return new KnotProxyVertxEBProxy(vertx, address);
   }
 
+  static KnotProxy createProxyWithOptions(Vertx vertx, String address, DeliveryOptions deliveryOptions) {
```
marcinczeczko: Same here as with AdapterProxy...

knotx-core/src/main/java/io/knotx/proxy/RepositoryConnectorProxy.java
```diff
@@ -31,5 +32,9 @@ static RepositoryConnectorProxy createProxy(Vertx vertx, String address) {
     return new RepositoryConnectorProxyVertxEBProxy(vertx, address);
   }
 
+  static RepositoryConnectorProxy createProxyWithOptions(Vertx vertx, String address, DeliveryOptions deliveryOptions) {
```
marcinczeczko: Same here as with AdapterProxy...

knotx-core/src/main/java/io/knotx/proxy/reactive/KnotProxyFactory.java
```diff
@@ -0,0 +1,36 @@
+/*
```
marcinczeczko: See my initial comment - factories are not required at all

knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java
Skejven: I'm just thinking out loud.
@marcinczeczko don't you agree that calling:
`AdapterProxy.createProxy(vertx, address, null)` looks odd?
Why the `null`? Should this interface force the user to pass `DeliveryOptions` parameter?
This is just to make the code look good, and skip in my opinion weird constructions like
`Factory.construct(param, null, null, param2)`.

knotx-knot/knotx-knot-action/src/main/java/io/knotx/knot/action/ActionKnotProxyImpl.java
```diff
@@ -102,7 +103,8 @@ private KnotContext handleGetMethod(List<FormEntity> forms, KnotContext knotCont
 
   private Single<AdapterResponse> callActionAdapter(KnotContext knotContext, FormEntity current) {
     LOGGER.trace("Process form for {} ", knotContext);
-    AdapterProxy adapter = AdapterProxy.createProxy(vertx, current.adapter().getAddress());
+    AdapterProxy adapter = AdapterProxyFactory
```
Skejven: Continuing @toniedzwiedz comment, this should rather look like:
```
AdapterProxy adapter = configuration.getDeliveryOptions().isPresent()
                                       ? AdapterProxyFactory.createProxy(vertx, configuration.getDeliveryOptions().get(), address)
                                       : AdapterProxyFactory.createProxy(vertx, address)
```

knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java
tomaszmichalak: You are right that Vert.x handle nulls. But as a result we will have to add nulls in all tests. So we can have AdapterProxy.createProxy(vertx, address, null) in all tests.

knotx-core/src/main/java/io/knotx/proxy/reactive/AdapterProxyFactory.java
tomaszmichalak: @toniedzwiedz Please check what is the expected behavior in this case. We have a configuration that has a getter:
```Optional<DeliveryOptions> get()```
It means that we expect DeliveryOptions may be not defined. Based on this value we want to create a factory method that uses two different methods with API (with and without DeliveryOptions). I don't know what is best in this situation...

File: knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java (hunk 2 above)
Comment (marcinczeczko) [reply knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java]: Yes, but I don't like overengineering by adding Factories to handle nulls, don't you think isn't too big for such small thing ?

Comment (Skejven) [reply knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java]: After the offline catchup with @marcinczeczko:
- let's not break compatibility and stay with 2 factory methods in `ProxyGen` helpers (like in this interface)
- classes like `XXXProxyFactory` are a bit overengineered, let's just construct Proxies directly where it is necessary
- `Optional` in config POJOs - suggestion here is to stick with Vert.x approach, and initialize not set properties with default values, e.g. in constructor:
```
...
deliveryOptions = config.containsKey("deliveryOptions") ? new DeliveryOptions(config.getJsonObject("deliveryOptions")) : new DeliveryOptions();
...
```

File: documentation/src/main/wiki/ActionKnot.md [hunk 7]
```diff
@@ -137,3 +137,24 @@ Adapter metadata options available. Take into consideration that Adapters are us
 | `params`                    | `JSON object`                       | &#10004;       | Default params which are sent to Adapter. |
 | `allowedRequestHeaders`     | `String`                            | &#10004;       | Array of HTTP client request headers that are allowed to be passed to Adapter. **No** request headers are allowed if not set. |
 | `allowedResponseHeaders`    | `String`                            | &#10004;       | Array of HTTP response headers that are allowed to be sent in a client response. **No** response headers are allowed if not set. |
+
+### Vert.x Event Bus delivery options
+
+While HTTP request processing, Action Knot calls Adapter using
```
marcinczeczko: `During HTTP request processing, Knot communicates with Adapter over the Vert.x event bus. Low level aspects of that communication can be configured using Vert.x Delivery Options object. The caller of the Adapter, such as Action or Service Knot, can customise default settings by providing a deliveryOptions field in the config section of its configuration.

Comment: Can we move a description of setting up delivery option to one place ? e.g. on Knot documentation page ? it would be hard to manage doc if it's in multiple places

tomaszmichalak:
CR fixes.

documentation/src/main/wiki/ActionKnot.md
```diff
@@ -137,3 +137,24 @@ Adapter metadata options available. Take into consideration that Adapters are us
...
+
+### Vert.x Event Bus delivery options
+
+While HTTP request processing, Action Knot calls Adapter using 
+[Vert.x Event Bus](http://vertx.io/docs/apidocs/io/vertx/core/eventbus/EventBus.html). The `config` field can contain 
+[Vert.x Delivery Options](http://vertx.io/docs/apidocs/io/vertx/core/eventbus/DeliveryOptions.html) related to the event 
+bus. It can be used to control the low level aspects of the event bus communication like timeouts, headers, message 
+codec names.
+
+The `deliveryOptions` need to be added in the following place, of the Action Knot configuration to define the
```
marcinczeczko: `deliveryOptions` needs to be added as shown below:


NOTE: don't use Action Knot here as it can be set on any Knot

documentation/src/main/wiki/Server.md
```diff
@@ -234,3 +235,37 @@ Other option is to provide those parameters through JVM properties:
...
+### Vert.x Event Bus delivery options
+
+While HTTP request processing, Server calls other modules like Repository Connectors, Knots using 
+[Vert.x Event Bus](http://vertx.io/docs/apidocs/io/vertx/core/eventbus/EventBus.html). 
+The `config` field can contain [Vert.x Delivery Options](http://vertx.io/docs/apidocs/io/vertx/core/eventbus/DeliveryOptions.html)
+related to the event bus. It can be used to control the low level aspects of the event bus communication like timeouts, 
+headers, message codec names.
+
+
+The `repositoryDeliveryOptions` need to be added in the following place, of the KnotxServerVerticle configuration to define
```
marcinczeczko: why it can be named deliveryOptions ??

documentation/src/main/wiki/ActionKnot.md
tomaszmichalak: Descriptions are similar, but not the same. I thought about that but we lose the information where (between what modules) the communication occurs.

documentation/src/main/wiki/Server.md
tomaszmichalak: We want to be able to specify deliveryOptions for a repository connector and knots separately.

documentation/src/main/wiki/ActionKnot.md
tomaszmichalak: @marcinczeczko The communication takes place in three places:
1. Between Server and Repository Connector / other Knots.
2. Between Service Knot and Service Adapters.
3. Between Action Knot and Action Adapters.

This is the reason we define those deliveryOptions in those places. So event publisher defines the event bus rules.

documentation/src/main/wiki/ActionKnot.md
tomaszmichalak: Done.

knotx-core/src/main/java/io/knotx/proxy/RepositoryConnectorProxy.java
tomaszmichalak: Done.

knotx-core/src/main/java/io/knotx/proxy/KnotProxy.java
tomaszmichalak: Done

knotx-core/src/main/java/io/knotx/proxy/AdapterProxy.java
tomaszmichalak: Fixed.

documentation/src/main/wiki/ActionKnot.md
tomaszmichalak: Fixed.

documentation/src/main/wiki/Server.md
Skejven: If we want to have separate delivery options per each connected address, maybe we should  have "global"  `deliveryOptions` that defines default options for each Proxy and additionally allow to define options inside each section and keep consistent name `deliveryOptions`, e.g.:

```json
{
  "main": "io.knotx.server.KnotxServerVerticle",
  "options": {
    "config": {
      "httpPort": 8092,
      "displayExceptionDetails": true,
      "deliveryOptions": {
        "timeout": 15000
      },
      "repositories": [
        {
          "path": "/content/local/.*",
          "address": "knotx.core.repository.filesystem",
          "deliveryOptions": {
            "timeout": 30000
          }
        },
        {
          "path": "/content/.*",
          "address": "knotx.core.repository.http",
          "deliveryOptions": {
            "timeout": 30000
          }
        }
      ],
      "splitter": {
        "address": "knotx.core.splitter"
      },
      "routing": {
        "GET": [
          {
            "path": ".*",
            "address": "knotx.knot.service",
            "deliveryOptions": {
              "timeout": 60000
            },
            "onTransition": {
              "next": {
                "address": "knotx.knot.handlebars"
              }
            }
          }
        ]
      },
      "assembler": {
        "address": "knotx.core.assembler"
      }
    }
  }
}
```

In the case above, following timeouts will be like:
- filesystem repository: 30s,
- http repository: 30s,
- knotx.knot.service 60s,
- all others 15s

tomaszmichalak: @Skejven Your proposal looks very interesting. I recommend to merge the current solution and investige more advanced possibilities if required in the future.

Skejven: Let's then get rid of `repositoryDeliveryOptions` and enable setting it with name `deliveryOptions` either on "globally" on verticle level (e.g. `Server`) or only on producer-level (e.g. only `repositories`).

PR #354, opened 2017-12-12
Feature/csrf

marcinczeczko:


knotx-server/src/main/java/io/knotx/server/configuration/KnotxCSRFConfig.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package io.knotx.server.configuration;
+
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.handler.CSRFHandler;
+
+
+public class KnotxCSRFConfig {
+
+  private long DEFAULT_TIMEOUT = 30 * 60 * 1000; // 30 minutes
```
Skejven: Can we make it constant (`static final`)?

knotx-server/src/main/java/io/knotx/server/configuration/KnotxCSRFConfig.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private String cookiePath;
+
+  private String headerName;
+
+  private String secret;
+
+  private long timeout;
+
+  public KnotxCSRFConfig(JsonObject config) {
```
Skejven: This class is really simple, but maybe we can use `@DataObject(generateConverter = true)` here?
Does it support default values?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ List of changes that are finished but not yet released in any final version.
  - [PR-350](https://github.com/Cognifide/knotx/pull/350) - Cosmetic cleanup of default logback.xml config files. Fixed URLs to the bootstrap CSS on example app.
  - [PR-352](https://github.com/Cognifide/knotx/pull/352) - Fix for #351 - Prevent NPE if service returns empty body
  - [PR-353](https://github.com/Cognifide/knotx/pull/353) - Fix for #351 - Deprecated [`ProxyHelper`](http://vertx.io/docs/apidocs/io/vertx/serviceproxy/ProxyHelper.html) replaced with [`ServiceBinder`](http://vertx.io/docs/apidocs/io/vertx/serviceproxy/ServiceBinder.html)
+ - [PR-354](https://github.com/Cognifide/knotx/pull/354) - Improvement - added CSRF protection and ability to configure it on any route.
```
Skejven: This looks like a feature that should be mentioned in `UpgradeNotes` so that after Knot.x upgrade, users should address csrf configuration.

knotx-server/src/main/java/io/knotx/server/configuration/KnotxServerConfiguration.java
```diff
@@ -23,6 +23,11 @@
 
 public class KnotxServerConfiguration {
 
+  //Default limit=-1 - Unlimited size
+  private final static long DEFAULT_UPLOAD_LIMIT = -1;
+
+  private final Long fileUploadLimit;
```
Skejven: Docs should be updated about new parameter.

knotx-server/src/main/java/io/knotx/server/configuration/KnotxCSRFConfig.java
marcinczeczko: done

knotx-server/src/main/java/io/knotx/server/configuration/KnotxServerConfiguration.java
marcinczeczko: fixed

knotx-server/src/main/java/io/knotx/server/configuration/KnotxCSRFConfig.java
marcinczeczko: I'd rather not go with data object generator now. It will make a incosistency. Ideally, all the configuration objects should be treated as DataObject - that will result in a great support for documentation purposes, but I it's something we should do globally for all configuration objects.

CHANGELOG.md
marcinczeczko: fixed

PR #357, opened 2018-01-04
Fixed #356 - Create adapter proxies in ServiceKnot on start instead o…

### LINKED ISSUE #356: Remember Proxy object reference
PROBLEM:
Proxy objects (KnotProxy and AdapterProxy) are created for all incoming requests. It causes hidden exceptions (IllegalStateExceptions). Those exceptions are not propagated down so Knot.x works correctly but it can influence the Knot.x performance. 
Those exceptions can be observed in profiler.

SOLUTION:
We should store references to Proxy objects once and use them while requests processing.

marcinczeczko:
…f for each request


tomaszmichalak:
Next changes here: https://github.com/Cognifide/knotx/pull/361

Issue #396, opened 2018-03-20
Summarize configuration update in wiki


Skejven:
### Context of the issue/feature

Summarize configuration update from #376, #397 and #402 in Upgrade Notes in order to notify the users of the changes that were introduced.

My thoughts:
Dokumentacja

- Getting started
   - How to start knotx with configs
   - From where to get configs 
   - Check that release will send only jars to maven central, configs should be taken manually from github (link to application, bootstrapo configs)
 - Debugging - to change
 - New section: Configuration
    - describe bootstrap.json - how to point application.conf, how to point multiple .conf files, how to point whole folder - how it works (merge all confs), you can still use json - show how, how to control enable/disable scanPeriod, describe how reload works (undeploy & deploy automatically)
    - describe application.conf, how it's structured, major sections, how to use include (consult hocon spec to properly describe includes), required() directive
    - sample config, show how to use cheatsheet.adoc to find option (if not available as comment), and how to set it up in .conf file
    - describe what errors are validated, e.g. if used required() on include fail knotx start if file does not exists, without required() it will ignore - optional include. Describe paramtere reference, e.g. ${global.address.xyz} that fails if wrong reference used

- On each module - do not list all config options - instead point to cheatsheet.adoc on github where all modules have described all available options. only describe how it works
- In modules where we're showing some example configs, e.g. SSL on server, routing, change those examples to HOCON format
- KnotxDeployment - a deployment should be like in starter, such as:
     - lib/ with all jars (knotx fat + any custom modules)
     - conf/ should have bootstrap.json, application.conf, logback.xml
     - java -jar command should set class path as follows -classpath="conf/:lib/*" so configs will be on classpath - then you don't need to provide path to logback config in command line

## My view on chunks of changes:
- [ ] Modify all md docs to use HOCON instead of JSON in any examples
- [ ] Modigy gettingStarted + KnotxDeployment in how to structure application (classpath, where to put configs etc. - starter like)
- [ ] Create new section: Configuration - explaining bootstrap and hocon and struture of the configurations
- [ ] The way how knotx is to be started - see #402

marcinczeczko:
Updated a ticket with the scope of the documentation changes.

marcinczeczko:
Documentation updated

PR #400, opened 2018-03-24
Reorganized standalone & example config files into hocon

marcinczeczko:
## Description
After migrating to HOCON form, a default configuration file for whole Knot.x is required. The config should have mentioned all configuration parameters with a comments describing its purpose.

## Motivation and Context
Existing config is missleading and user need to consult multiple places to find possible options to configure

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ ] I have updated the documentation accordingly. Documentation to be covered in #396 

---


knotx-example/knotx-example-app/config/application.conf
```diff
@@ -9,63 +10,107 @@ modules = [
...
 
-repo {
-  domain = localhost
-  port = 3001
-}
+########### Globals ###########
+global {
+  serverPort = 8092
+
+  snippetTagName = script
```
tomaszmichalak: Maybe it is the time to introduce the custom tag name like "knotx:snippet" :) ?

knotx-standalone/config/application.conf
```diff
@@ -1,12 +1,103 @@
...
-    "serviceAdapter=io.knotx.adapter.service.http.HttpServiceAdapterVerticle"
-  ]
+  "server=io.knotx.server.KnotxServerVerticle"
+  "httpRepo=io.knotx.repository.http.HttpRepositoryConnectorVerticle"
+  "splitter=io.knotx.splitter.FragmentSplitterVerticle"
+  "assembler=io.knotx.assembler.FragmentAssemblerVerticle"
+  "hbsKnot=io.knotx.knot.templating.HandlebarsKnotVerticle"
+  "serviceKnot=io.knotx.knot.service.ServiceKnotVerticle"
+  # Enable action knot if you want to have support for form submittions. Follow the actionKnot configuration to set it up
+  # "actionKnot=io.knotx.knot.action.ActionKnotVerticle"
```
tomaszmichalak: Action Knot is disabled but configured in the routing.

knotx-example/knotx-example-app/config/application.conf
Skejven: We definitely have to change the default to e.g. `knotx:snippet`. But not yet, let's not break backward compability.

marcinczeczko: I agree with skejven. Making it default to something else, not only break the backward compatibility, but also introduces a lot of refactors of tests - that anyway need to be rebuilt, changed, reorganized...

knotx-standalone/config/application.conf
marcinczeczko: Yes, I'm thinking actually about enabling actionKnot and leaving the config as it's now - and clarify in the config that if you're not using any module, just comment it out or remove.

knotx-example/knotx-example-app/config/includes/serviceKnot.conf
```diff
@@ -0,0 +1,31 @@
+# Event bus address of the handlebars Knot
+address = ${global.address.knot.service}
+
+# An
```
Skejven: Some missing comment here?

knotx-example/knotx-example-app/config/includes/serviceAdapter.conf
```diff
@@ -0,0 +1,59 @@
+# Event bus address of the Basic HTTP Service Adapter
```
Skejven: Do we need here an example (commented out) how to setup connection via SSL?

marcinczeczko: Added more stuff

PR #404, opened 2018-04-04
Refactor Knotx Launcher and the way how it's started

marcinczeczko:
## Description
Changed:
- removed KnotxLauncher and used vertx OOTB
- instead created vertx cli command to start knotx
- expose only two commands for knotx users: run-knotx, bare (for high-availability), other vertx ootb commands does not make sense at this moment
- added config for fs repo in example
- small cleanup in logback.xml to not pollute console/logs of logback status lines
- added default cluster configuration - that might be used to show clustering & ha
- prepare right project structure that will be compatible with the starter project

## Motivation and Context
- general mess and too many hacks :)
- cleaner way of knotx starting
- ability to add own commands by the project if needed

## Types of changes

## Checklist:

---


documentation/src/main/wiki/GettingStarted.md
```diff
@@ -26,7 +26,7 @@ First download Knot.x sample app & config for latest version, or build it yourse
 
 Now you can run Knot.x:
 ```
-java -Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory -Dlogback.configurationFile=config/logback.xml -jar knotx-example-app-X.Y.Z-fat.jar -conf config/bootstrap.json
+java -Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory -cp config:target/knotx-example-app-X.Y.Z-fat.jar io.vertx.core.Launcher
```
Skejven: Shouldn't be here `... -cp config:lib/* io.vertx.core.Launcher` ?
And I believe few lines before we should define to download `knotx-example-app-X.Y.Z.fat.jar` into `lib` directory.

knotx-core/src/main/java/io/knotx/launcher/KnotxCommand.java
```diff
@@ -0,0 +1,215 @@
...
+  private static final int KNOTX_MISSING_OR_EMPTY_CONFIGURATION_EXIT_CODE = 30;
+
+  private static final String DEFAULT_LOADER_FILE = "bootstrap.json";
+
+
+  private DeploymentOptions deploymentOptions;
+
+  private String config;
+
+  private boolean ha;
```
tomaszmichalak: We need to provide the description which covers both `BareCommand` options.

knotx-core/src/main/java/io/knotx/launcher/KnotxCommand.java
```diff
@@ -0,0 +1,215 @@
...
+   * Starts vert.x and deploy the verticle.
+   */
+  @Override
+  public void run() {
+    JsonObject conf = getConfiguration();
+    if (conf == null) {
+      ExecUtils.exit(KNOTX_MISSING_OR_EMPTY_CONFIGURATION_EXIT_CODE);
+    }
+
+    super.run();
```
tomaszmichalak: Not sure if it is the correct solution. Please check `io.vertx.core.impl.launcher.commands.BareCommand#afterStartingVertx` method which is called when vert.x instance is set - those hooks should happen after Knot.x verticles deployment. In the current solution those hooks happen before event Knot.x Starter Verticle run.

PR #407, opened 2018-04-10
Feature/main config location

marcinczeczko:
## Description
Added vertx hooks to properly terminate instance on fatal failure, like missing configurations etc.

## Motivation and Context
- During testing it appeared that Knotx is not shutting down when invoked System.exit() methods, it's due to the fact vertx create shutdown hook that executes proper hooks on each command
- Additionally, added better error control in starter knotx to stop instance for multiple scenarios 
- Added support for `KNOTX_HOME` variable in bootstrap.json, to better control location of config files, especially in dockerized environments

## Types of changes

## Checklist:

---

### DIFF
--- a/knotx-core/src/main/java/io/knotx/util/StringUtil.java
+++ b/knotx-core/src/main/java/io/knotx/util/StringUtil.java
@@ -1,0 +1,36 @@
+package io.knotx.util;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class StringUtil {
+
+  /**
+   * Returns the string value of the system property with the
+   * specified name. The first argument is treated as the name of a
+   * system property.  System properties are accessible through the
+   * {@link java.lang.System#getProperty(java.lang.String)} method.
+   *
+   * <p>The second argument is the default value. The default value is
+   * returned if there is no property of the specified name or if the
+   * specified name is empty or {@code null}.
+   *
+   * @param propertyName property name.
+   * @param defaultVal default value.
+   * @return the {@code String} value of the property.
+   * @throws SecurityException for the same reasons as {@link System#getProperty(String) System.getProperty}
+   * @see System#getProperty(java.lang.String)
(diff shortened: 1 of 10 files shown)


knotx-core/src/main/java/io/knotx/util/StringUtil.java
```diff
@@ -0,0 +1,52 @@
...
+   * specified name is empty or {@code null}.
+   *
+   * @param propertyName property name.
+   * @param defaultVal default value.
+   * @return the {@code String} value of the property.
+   * @throws SecurityException for the same reasons as {@link System#getProperty(String) System.getProperty}
+   * @see System#getProperty(java.lang.String)
+   * @see System#getProperty(java.lang.String, java.lang.String)
+   */
+  public static String getString(String propertyName, String defaultVal) {
```
Skejven: Maybe `getSystemProperty` would be a better name here?

knotx-core/src/main/java/io/knotx/util/StringUtil.java
```diff
@@ -0,0 +1,52 @@
...
+    String value = null;
+    try {
+      value = System.getProperty(propertyName);
+    } catch (IllegalArgumentException | NullPointerException e) {
```
Skejven: Shouldn't we log it?

knotx-core/src/main/java/io/knotx/util/StringUtil.java
marcinczeczko: I followed the Java convention, see `java.lang.Integer.getInteger(nm,dflt)` or Double.getDouble()...

knotx-core/src/main/java/io/knotx/util/StringUtil.java
marcinczeczko: There is no need to do it. The whole idea is to get a nice fallback method - if sys prop is not defined use default value - no value to log it at all - even Java Integer.getInteger swallows it

PR #409, opened 2018-04-15
Refactor StringToPatternFunction and add regex escaping

piandrus:
## Description

The StringToPatternFunction was cleaned a bit. Now it has it's own unit tests. The pattern now includes escaping of regex constructs, so only "*" works.

## Motivation and Context

I'm trying to work my way inside-out to get familiar with the project. This class seemed like a good start. 

I noticed that user could pass some regex constructs to it and they would be included in the resulting pattern, so I added escape characters.

I'm not sure if it's intended, but the "*" wildcard does not work as I was expecting - it matches at least one character. Is that correct?

## Types of changes

## Checklist:

---

### DIFF
--- a/knotx-core/src/main/java/io/knotx/http/StringToPatternFunction.java
+++ b/knotx-core/src/main/java/io/knotx/http/StringToPatternFunction.java
@@ -27,12 +27,19 @@ public class StringToPatternFunction implements Function<String, Pattern> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(StringToPatternFunction.class);
 
-  private static final String WILDCARD = "*";
+  private static final StringToPatternFunction INSTANCE = new StringToPatternFunction();
 
+  private static final String WILDCARD = "*";
   private static final String REGEX_START_ESCAPING = "\\Q";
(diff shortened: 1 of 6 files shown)


knotx-core/src/main/java/io/knotx/http/StringToPatternFunction.java
```diff
@@ -26,21 +27,38 @@
...
+  }
+
+  private Pattern makePattern(String stringPattern) {
+    String regex = toRegex(stringPattern);
+    return Pattern.compile(regex, Pattern.CASE_INSENSITIVE);
   }
 
   private String toRegex(String stringPattern) {
-    return "^" + stringPattern.replace(WILDCARD, "(.+)") + "$";
+    return "^" + REGEX_START_ESCAPING
```
Skejven: We actually want to have other regexp functionalities in headers definitions.

piandrus:
Thank you @Skejven for the review!
I left some non-functional modifications in the `StringToPatternFunction`, but I can revert them as well if you wish.

Skejven:
Thank you @piandrus for updating this Pull Request, that's great.

PR #385, opened 2018-02-06
Support for snippet tags other than script

Skejven:
## Description
Support for snippet tags other than `script`.

## Motivation and Context
There should be more freedom when it comes to defining dynamic Knot.x snippets. Current solution supports only snippets defined inside `<script>` tag. While it suits most of the cases there would be few that requires change of the snippet tag name to some custom one because of how Knot.x finds snippets (Knot.x just scans the markup for the opening of snippet tag and the first occurrence of end of the tag).
You may find more motivation in #107 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the [code style](https://github.com/Cognifide/knotx/blob/master/CONTRIBUTING.md#coding-conventions) of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

---
I hereby agree to the terms of the Knot.x Contributor License Agreement.

PR #419, opened 2018-04-26
Custom parameters prefix knotx snippet

Skejven:
## Description
Snippet parameters names are too long, because of being HTML5 compliant (as the snippet tag is `<script>`). Since Knot.x `1.2.1`, it is possible to customize snippet tag name (see #385 for more details).
This enchancment is followup of [this discusion](https://github.com/Cognifide/knotx/pull/385/files#r183280005).

## Motivation and Context
- snippet parameters should enable defining HTML compliant prefixes, but when not necessary should be simplified.

## Types of changes

## Checklist:

---


Skejven:
@pun-ky thanks for this catch, of course it is not required.
I've updated wiki with information about it.
Good catch.

knotx-core/src/main/java/io/knotx/options/SnippetOptions.java
```diff
@@ -0,0 +1,100 @@
...
+public class SnippetOptions {
+
+  private String tagName;
+  private String paramsPrefix;
+
+  /**
+   * Default constructor
+   */
+  public SnippetOptions() {
+  }
```
marcinczeczko: what about default values ? a good practice is to have private init() method invoked in default constructor and inside jsonobject constructor (before converter). an init method will simply set default values for tag and prefix. that way we avoid potential nullpointers

PR #421, opened 2018-05-14
Value resolver for hocon that converts the sys prop variable into int…

marcinczeczko:
## Description
A HOCON value resolver that's capable of doing value type detection. It should be able to detect Integer, Boolean and finally string based on the value. It's a must have in order to have an ability to merge with JSON having values in those types

## Motivation and Context
HOCON supports variables resolving from system properties. However, by design it always wraps it as String. However, we're using JSON extensively to configure verticles. Verticle implementations are looking forward for values in type of Integer (e.g. port) or boolean, so there is a need to be able to properly inject those values.

## Types of changes

## Checklist:

---


knotx-core/src/main/java/io/knotx/launcher/config/ConfProcessor.java
```diff
@@ -81,6 +90,33 @@ public void process(Vertx vertx, JsonObject configuration, Buffer input,
...
+    private final Properties sysProps = System.getProperties();
+
+    @Override
+    public ConfigValue lookup(String path) {
+      String value = sysProps.getProperty(path);
+      if (StringUtils.isNotBlank(value)) {
+        if (NumberUtils.isCreatable(value)) {
+          return ConfigValueFactory.fromAnyRef(NumberUtils.toInt(value));
+        }
+        if (Boolean.TRUE.toString().equalsIgnoreCase(value) || Boolean.FALSE.toString().equalsIgnoreCase("false")) {
```
Skejven: This statement will always return true because of `Boolean.FALSE.toString().equalsIgnoreCase("false")` ;)

marcinczeczko:
fixed comment

documentation/src/main/wiki/Configuration.md
```diff
@@ -131,4 +131,21 @@ clientOptions {
 
 ```
 In rare situation you can always consult the [Knot.x verticles cheatsheet](https://github.com/Cognifide/knotx/blob/master/documentation/src/main/cheatsheet/cheatsheets.adoc)
-to find all available configuration options.
\ No newline at end of file
+to find all available configuration options.
+
+### System properties
```
Skejven: Maybe this information is worth to be mentioned in changelog and upgrade notes?

PR #422, opened 2018-05-28
Custom handlebars delimiters

marcinczeczko:
## Description
There are use cases in which the template to be rendered by Knot.x consists of handlebars scripts that are to be interpreted client side e.g. using Angular. In order to do so, an Angular variables, expressions must not be processed by the Knot.x.
In order to enable this, there is a need to have Handlebars delimiters configured so, the developer might chose to use `[<` and `]>' delimiters instead of `{{` and `}}`

## Motivation and Context
- To split the HBS script between server-side and client side

## Types of changes

## Checklist:

---

### DIFF
--- a/documentation/src/main/wiki/HandlebarsKnot.md
+++ b/documentation/src/main/wiki/HandlebarsKnot.md
@@ -49,9 +49,10 @@ Handlebars Knot uses data from Fragment Context and applies it to Fragment Conte
 Finally Fragment Content is replaced with merged result.
 
 ## Interpolation symbol
-By default Handlebars engine uses `{{` and `}}` symbols as delimiters of variables, expressions, etc.
-However, since the Knot.x can be used to generate markup on server side, it might also consists of markup to be interpreted by the front end engine, such as Angular.
-To avoid conflicts between mustache script to be executed server side from the client side, a Handlebars knot introduces two configuration parameters that enables you to configure custom symbols to be used in Knot.x snippets.
+By default , the Handlebars engine uses `{{` and `}}` symbols as tag delimiters.
+However, while Knot.x can be used to generate markup on the server side, the very same page might also contain templates intended for client-side processing. 
+This is often the case when frameworks like Angular.js or Vue.js are used.
+To avoid conflicts between Mustache templates to be executed server-side and ones evaluated on the client side, a Handlebars knot introduces two configuration parameters that enable you to configure custom symbols to be used in Knot.x snippets.
 E.g.:
 In order to use different symbols as below
 ```html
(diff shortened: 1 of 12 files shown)


knotx-it-tests/src/test/java/io/knotx/example/monolith/SampleApplicationIT.java
```diff
@@ -83,6 +83,12 @@ public void whenRequestingLocalSimplePageWithGet_expectLocalSimpleHtml(TestConte
     testGetRequest(context, LOCAL_REQUEST_URI, "localSimpleResult.html");
   }
 
+  @Test
+  @KnotxConfiguration("knotx-test-app-custom-symbol.json")
```
Skejven: It may be worth to add one additional tests - set custom symbol, but also use the default in the processed snippet markup. Expect custom symbols processed, but `{{` and `}}` left for latter processing by any other scripts/engines (e.g. angular).

marcinczeczko: Fixed

documentation/src/main/wiki/HandlebarsKnot.md
```diff
@@ -48,6 +48,25 @@ Handlebars Knot uses data from Fragment Context and applies it to Fragment Conte
 ```
 Finally Fragment Content is replaced with merged result.
 
+## Interpolation symbol
+By default Handlebars engine uses `{{` and `}}` symbols as delimiters of variables, expressions, etc.
+However, since the Knot.x can be used to generate markup on server side, it might also consists of markup to be interpreted by the front end engine, such as Angular.
```
toniedzwiedz:
```
By default Handlebars engine uses `{{` and `}}` symbols as delimiters of variables, expressions, etc.
However, since the Knot.x can be used to generate markup on server side, it might also consists of markup to be interpreted by the front end engine, such as Angular.
```
Just _Knot.x_, _the_ is superfluous. I understand you mean Knot.x as a tool, not a specific instance of Knot.x
As a matter of fact, the article _the_ could be used in the line before, when referring to _the Handlebars engine_ (that Knot.x uses).

How about:
```
By default , the Handlebars engine uses `{{` and `}}` symbols as tag delimiters.
However, while Knot.x can be used to generate markup on the server side, the very same page might also contain templates intended for client-side processing. This is often the case when frameworks like Angular or Vue.js are used.
```

documentation/src/main/wiki/HandlebarsKnot.md
```diff
@@ -48,6 +48,25 @@ Handlebars Knot uses data from Fragment Context and applies it to Fragment Conte
...
+To avoid conflicts between mustache script to be executed server side from the client side, a Handlebars knot introduces two configuration parameters that enables you to configure custom symbols to be used in Knot.x snippets.
```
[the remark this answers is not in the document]
marcinczeczko: Thanks for your comments. Fixed

documentation/src/main/wiki/HandlebarsKnot.md
marcinczeczko: Thanks for your comments. Fixed

Issue #391, opened 2018-02-28
Static data/params on the Snippet level available to the handlebars 


marcinczeczko:
### Version used

* knot.x: 1.2.1

### Context of the issue/feature
Snippet fragment sometime should use some data that were processed during rendering of the <script> tag, by modifying HTML structure. However, since the knotx snippet is in the special HTML tag, some CMS system are not able to process the tags inside the fragment.
In such a cases, CMS could render this data (e.g. paths to image renditions into the snippet param attribute. Data from such attribute would be available in the Handlebars context, so the data can be easily rendered.
```html
<script data-knotx-knots="services,handlebars"
            data-knotx-service="some-service"
            data-knotx-service-data="{'img':'/path/image.png'}"
            type="text/knotx-snippet">
    <img src="{{_result._data.img}}"/>
</script>
```
A new attribute "data-knotx-service-data". Other approach, is to expose data from data-knotx-service-params attribute... into the handlebars context.

### Bug reproducer

None

### Steps to reproduce

None

### Appendix

None

tomaszmichalak:
I thought about it by analyzing the isolation of nodes in the routing (Knots). My idea was to have some kind of namespace for each Knot.

Example:
```html
<script data-knotx-knots = "services, steering wheels"
    data-knotx-service = "some-service"
    data-knotx-service-params = "{'img': '/ path / image.png'}"
    data-knotx-handlebars-data = "{'img': '/ path / image.png'}"
    type = "text / knotx-snippet">
        <img src = "{{_ result._data.img}}" />
</ script>
```
Then the Service Knot sees only params within its namespace (service), the same Handlebars Knot does.

tomaszmichalak:
In this way we can have some global params: `data-knotx-all-param1 = "some-service"`.

Skejven:
Won't fix for #107 from 1.2.1 help this problem (change the `<script>` tag to something other).
I'm not sure about this idea... If we want to go this way we shouldn't mix `_result` with some static data that is actually not from the response.

```xml
<script data-knotx-knots="services,handlebars"
            data-knotx-service="some-service"
            data-knotx-service-const="{'img':'/path/image.png'}"
            type="text/knotx-snippet">
    <p>Hello {{_result.message}}</p>
    <img src="{{_const.img}}"/>
</script>
```

PR #426, opened 2018-06-29
static fragment data passed to handlebars context

scypio:
Static fragment data passed to the handlebars. 

## Description

During FragmentContext initialization static data (prefixed with `data-const-` is now passed stored into the context. Prefix used for these properties is `_const`. Therefore data attribute `data-const-foo` can be referenced from handlebars as `{{_const.foo}]`

## Motivation and Context

Better code organization - and solution for some CMS systems that are not able to process the tags inside the fragment.

Implementation of https://github.com/Cognifide/knotx/issues/391

## Types of changes

## Checklist:

---

### DIFF
--- a/knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/impl/FragmentContext.java
+++ b/knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/impl/FragmentContext.java
@@ -20,10 +20,14 @@ import io.knotx.dataobjects.Fragment;
 import io.knotx.knot.service.service.ServiceAttributeUtil;
 import io.knotx.knot.service.service.ServiceEntry;
 import io.reactivex.Observable;
+import io.vertx.core.json.JsonObject;
+
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
 import org.jsoup.Jsoup;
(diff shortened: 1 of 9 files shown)


scypio:
@Skejven changelog and upgrade notes updated :)

knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/impl/FragmentContext.java
```diff
@@ -33,6 +37,7 @@
 
   private static final String DATA_SERVICE = ".*service.*";
   private static final String DATA_PARAMS = ".*params.*";
+  private static final String DATA_CONSTANTS = ".*const.*";
```
tomaszmichalak: It should not be a part of Service Knot in my opinion.

knotx-knot/knotx-knot-service/src/main/java/io/knotx/knot/service/impl/FragmentContext.java
```diff
@@ -100,6 +115,12 @@ private FragmentContext services(List<ServiceEntry> services) {
     this.services = services;
     return this;
   }
+  
+  private FragmentContext applyConstants(Map<String, String> constants) {
```
tomaszmichalak: This change should be moved to Knot.x Data Bridge.

tomaszmichalak:
This PR is outdated. Please check Knot.x 2.0 implementation to validate this idea.

Issue #423, opened 2018-06-04
Throw WARN/INFO instead of ERROR on 404


karoldrazek:
Please kindly consider throwing WARN or INFO rather than ERROR when there's 404 being thrown from repository.

Relevant code part:
https://github.com/Cognifide/knotx/blob/master/knotx-core/src/main/java/io/knotx/repository/http/HttpRepositoryConnectorProxyImpl.java#L151

toniedzwiedz:
```
    if (httpResponse.statusCode() >= 300 && httpResponse.statusCode() < 400) { //redirect responses
      LOGGER.info("Repository 3xx response: {}, Headers[{}]", httpResponse.statusCode(),
          DataObjectsUtil.toString(httpResponse.headers()));
    } else if (httpResponse.statusCode() != 200) {
      LOGGER.error("Repository error response: {}, Headers[{}]", httpResponse.statusCode(),
          DataObjectsUtil.toString(httpResponse.headers()));
    }
```

I'm not sure I understand this correctly but doesn't this effectively treat a number of 2xx family status codes as errors?

As for @karoldrazek 's suggestion, perhaps the log level could be different for 4xx  codes (warn/info) and for 5xx codes (error), the rationale being that the 400s can usually be fixed by the client by changing the request accordingly, whereas the 500s are server-side errors beyond the client's control.

marcinczeczko:
@toniedzwiedz check the condition, the response is logged as error if status code is different than `200`.
Anyway, this method is a great candidate for someone who'd like to contribute first time. In my opinion that `if` statement should log the repository responses in a following way:
- If status code is `2xx` log on `debug` level
- If status code is `3xx` log as `info`
- If status code is `4xx` or `5xx` log as `warn` - as it's not an application error but rather something unexpected that should warn the operator

And small tip. Instead of checking ranges of status codes, you can use a `HttpStatusClass` enum that simplifies response code checking, e.g., in order to test if the `500 =< code < 600` use:
```java
HttpStatusClass.SERVER_ERROR.contains(httpResponse.statusCode());
```

michalwronowski:
Logging `5xx` on `error` level as @toniedzwiedz suggested seems very reasonable.

PR #427, opened 2018-06-29
Response code logging tweaks - different log levels for success, redi…

scypio:
…rect and error responses

Currently every response from repository that return HTTP status code different than 200 is logged using ERROR logger level. The strategy has been adjusted.

## Description

2xx - success - log at debug
3xx - redirect - log at info
4xx - client error - log at warn
5xx - server error - log at error
other - log at warn 

https://github.com/Cognifide/knotx/issues/423

## Motivation and Context

Too many error messages may trigger unwanted warnings for devops. Error log level should indicate that something is wrong. 

## Types of changes

## Checklist:

---

### DIFF
--- a/knotx-core/src/main/java/io/knotx/repository/http/HttpRepositoryConnectorProxyImpl.java
+++ b/knotx-core/src/main/java/io/knotx/repository/http/HttpRepositoryConnectorProxyImpl.java
@@ -156,7 +156,7 @@ public class HttpRepositoryConnectorProxyImpl implements RepositoryConnectorProx
     } else if (HttpStatusClass.CLIENT_ERROR.contains(statusCode)) { // errors
       LOGGER.warn("Repository client error 4xx response: {}, Headers[{}]",
           statusCode, DataObjectsUtil.toString(httpResponse.headers()));
-    } else if (HttpStatusClass.SERVER_ERROR.contains(httpResponse.statusCode)) {
+    } else if (HttpStatusClass.SERVER_ERROR.contains(statusCode)) {
       LOGGER.error("Repository server error 5xx response: {}, Headers[{}]",
           statusCode, DataObjectsUtil.toString(httpResponse.headers()));
     } else {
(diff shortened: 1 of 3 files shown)


Skejven:
Thank you @scypio for this fix.

knotx-core/src/main/java/io/knotx/repository/http/HttpRepositoryConnectorProxyImpl.java
```diff
@@ -144,12 +145,21 @@ private String encode(String value) {
   }
 
   private ClientResponse toResponse(Buffer buffer, final HttpClientResponse httpResponse) {
-    if (httpResponse.statusCode() >= 300 && httpResponse.statusCode() < 400) { //redirect responses
-      LOGGER.info("Repository 3xx response: {}, Headers[{}]", httpResponse.statusCode(),
+    if (HttpStatusClass.SUCCESS.contains(httpResponse.statusCode())) {
```
marcinczeczko: For a better readability of each condition, get the `.statusCode()` into variable and do the check against that variable

scypio: Done

Skejven: I think if it is not useful information to see the request path (especially in `warn` and `error` level) in the message.
What do you think?

mateuszgrab-cognifide:
Would it be possible to add information about requested resource while logging response?
The issue I have is that it makes no sense in some cases to log everu request, however it is good to know what resource was requested if there was >= 400 response code.

Skejven:
What do you think about adding request path (especially in warn and error level) in the message?
I think that might be useful.

scypio:
@Skejven @mateuszgrab-cognifide I added request URL logging for 4xx, 5xx response codes

